PR #8, opened 2021-05-13
[simple-app] Make the chart "connection" test configurable

diranged:
**What did I want?**

The existing test-case was hard coded to a plain-text HTTP get with `wget` as the client. That doesn't make it easy to support alternative use-cases (gRPC for example).

**What did I do?**

Reconfigured the test suite so that all of the details are in the `values.yaml` file under the `tests.connection` key. This not only makes the test more configurable, but it also lets us start adding in other tests down the road that might be different and gives us independent control to configure them.

### DIFF
--- a/ct.yaml
+++ b/ct.yaml
@@ -4,5 +4,3 @@ chart-dirs:
   - charts
 helm-extra-args: --timeout 600s
 target-branch: main
-excluded-charts:
-  - iam-manager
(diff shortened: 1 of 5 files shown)


ct.yaml
```diff
@@ -4,5 +4,3 @@ chart-dirs:
   - charts
 helm-extra-args: --timeout 600s
 target-branch: main
-excluded-charts:
```
diranged, line 7: This was just left-over junk.

PR #9, opened 2021-05-17
[local-kafka] An opinionated local development kafka stack

diranged:
**What did I want?**

The existing `lightstream` and `featurestore` repositories both implement their own `make kafka` type Make target to set up a local Kafka instance. I believe it is simpler and more correct for us to use a common "local kafka chart" that can be sourced as a Chart Dependency by many future projects. I wanted to create this chart to reduce duplication, and encourage full end-to-end test suites of our helm charts using the `ct` test tool.

**What did I do?**

I have created a new chart named `local-kafka` (to be followed up by charts like `local-redis`, `local-elasticsearch` and more). The aim of the chart is to launch Kafka in a similar-to-production way using the Strimzi Operator and proper SSL-based authentication - but to make it really simple for local dev ... a simple `user` secret with the creds, no futzing around with ACLs (just `*` privileges), and no need for a large redundant kafka installation locally.

The chart is fairly customizable - but the goal is to _not_ customize it most of the time and simply use it as is.

**What needs to be done?**

After this is merged and begins being used by developers, we should fill out the `README.md.gotmpl` file with a full example of how to use the chart for real. We may, at that point, also want to make the `templates/tests/*` files optional.


diranged:
@sabw8217 @PiercingDan, I think this is ready for an initial commit.. thoughts?

PR #10, opened 2021-05-20
[simple-app]: Add in volumes/volumeMounts parameters

diranged:


charts/simple-app/templates/ingress/ingress.yaml
```diff
@@ -12,16 +12,6 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  {{- if .Values.ingress.tls }}
```
diranged, line 15: Removed this boilerplate because we don't use this anywhere - its specific to the `nginx` ingress controller.

PiercingDan (approved):

PR #11, opened 2021-05-21
[simple-app] Completely refactor the ingress/virtualService handling

diranged:
**What did I want?**

The existing `Ingress` and `VirtualService` handling was confusing and inconsistent. I wanted to
simplify it - by making an opinionated decision about the "Standard" resource configuration that
we should use.

**What did I do?**

1. Refactored the `Ingress` template so that the HTTP rules were hard-coded rather than using
    big nested strings that are then run through `tpl`. The new opinionated framework more closely
    matches what we do in our normal ALBs. This also just makes the template _much_ easier to
    understand and document. If a user needs a custom Ingress, they can create it on their own.

2. Refactored the `VirtualService` template to also be highly opinionated and favor a very simple
    single-service routing model. Custom `VirtualService` resources can be defined by the user
    if they need one.

3. Renamed the `ingressGateway` top level values key to `virtualService` to more accurately
    reflect the resource that it is configuring.

4. Removed unnecessary `service.port` parameters because we use the `ports[]` list instead
    anyways. Even I forgot what that value was being used for (hint: it was just used for a test)
    so it was clearly confusing.

5. Improved the test-suite by adding in some handling of connection-refused errors - this is
    because the test-container starts up faster than the service container does, and we were
    seeing random race condition failures.

### DIFF
--- a/charts/simple-app/templates/istio/virtualservice.yaml
+++ b/charts/simple-app/templates/istio/virtualservice.yaml
@@ -1,25 +1,36 @@
-{{- if .Values.ingressGateway.enabled }}
+{{- if .Values.virtualService.enabled }}
+{{- $global := . }}
 apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
 metadata:
   name: {{ include "simple-app.fullname" . }}
   labels:
     {{- include "simple-app.labels" . | nindent 4 }}
-  {{- with .Values.ingressGateway.annotations }}
+  {{- with .Values.virtualService.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  hosts:
-    {{- tpl .Values.ingressGateway.hosts . | nindent 4 }}
   gateways:
-    - {{ .Values.ingressGateway.namespace }}/{{ .Values.ingressGateway.gateway }}
-  {{- if .Values.ingressGateway.http }}
+    - {{ .Values.virtualService.namespace }}/{{ .Values.virtualService.gateway }}
+  hosts:
+    {{- range .Values.virtualService.hosts }}
+    - {{ tpl . $global }}
+    {{- end }}
+  {{- /* https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRoute */}}
   http:
-    {{- tpl .Values.ingressGateway.http . | nindent 4 }}
-  {{- end }}
-  {{- if .Values.ingressGateway.tls }}
+    - match:
+      {{- /* https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPMatchRequest */}}
+      - uri:
+          prefix: {{ .Values.virtualService.path }}
+      {{- /* https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRouteDestination */}}
+      route:
+        - destination:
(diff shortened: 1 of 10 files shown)


charts/simple-app/values.yaml
```diff
@@ -228,10 +260,10 @@ affinity: {}
 tests:
   connection:
     # -- The command used to trigger the test.
-    command: ['curl']
+    command: ['curl', '--retry-connrefused', '--retry', '5']
```
mrterry, line 263: do you want to add a retry delay?

diranged: There is an automatic doubling backoff delay - so it goes 1s, 2s, 4s, 8s, 16s...

charts/simple-app/values.yaml
```diff
@@ -228,10 +260,10 @@ affinity: {}
...
     # -- A list of arguments passed into the command. These are run through the tpl function.
     args:
-      - '{{ include "simple-app.fullname" . }}:{{ .Values.service.port }}'
+      - '{{ include "simple-app.fullname" . }}'
```
mrterry, line 266: this is reflecting the decision that your container runs on port 80, plaintext, & ssl termination happens in a sidecar.  corrrect?

charts/simple-app/values.yaml
```diff
@@ -51,7 +55,7 @@ proxySidecar:
     tag: latest
 
     # -- (String) Always, Never or IfNotPresent
-    pullPolicy: Always
+    pullPolicy: IfNotPresent
```
mrterry, line 58: does `IfNotPresent` cause problems with a deploy?  hopefully we're using uniquely tagged containers for deployments, but if someone was using `:latest` this would not necessarily get updates.

charts/simple-app/values.yaml
diranged: Yes - though more accurately, this is reflecting the _exact case described in the simple-app values.yaml file_.. which is a simple nginx service listening on port 80.

charts/simple-app/values.yaml
diranged: This setting is for the optional `proxySidecar` - the main application image is already set to use `IfNotPresent`. These are overridable of course - but in general I find the `Always` pull practice to be bad because it just slows down the deploy process.

charts/simple-app/templates/istio/virtualservice.yaml
```diff
@@ -1,25 +1,36 @@
-{{- if .Values.ingressGateway.enabled }}
+{{- if .Values.virtualService.enabled }}
+{{- $global := . }}
```
mrterry, line 2: i have no idea what this is actually doing.

mrterry (commented):
lgtm, but I don't have the most context.

diranged:
For a bit of proof, here's how the _final_ diff really doesn't change much for a real app: https://github.com/Nextdoor/nd-gogo/pull/12

charts/simple-app/values.yaml
```diff
@@ -143,55 +147,83 @@ securityContext: {}
 
 service:
   type: ClusterIP
-  port: 80
```
mrterry, line 146: Why did you remove this?

diranged: This was a misleading setting - it was only being used by the test framework.

PR #14, opened 2021-06-03
[istio-endpoints] Create a chart for managing per-namespace Istio configurations

diranged:
**What did I want?**

Managing Istio configurations is complicated and risky. I want to centralize the management of the configurations for different service-types so that we can have some consistency.

**What did I do?**

I have created a new `istio-endpoints` chart that manages the entire Istio sidecar configuration for a given Namespace. The motivation here is to not only allow developers to use this chart directly, but to also use this chart in the `simple-app` as an optional dependency chart.


mrterry:
lgtm, but this is beyond my knowledge base to review

charts/istio-endpoints/README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+### Set up your list of ElastiCache AWS Endpoints
+
+```yaml
+# values.yaml
+istio-endpoints:
+  test:
+    address: staging-cluster.abcd8x.clustercfg.usw2.cache.amazonaws.com
+    bindPort: 1234          # required
```
mrterry, line 69: how do I reference isto-endpoints.test.address in my application `{{ .Values.istio-endpoints.test.address }}`? or do I have to go the `global` thing?

mrterry (approved):
i think this makes sense

charts/istio-endpoints/README.md
diranged: Your application actually points to the `bindAddress` and `bindPort` that you have configured here, so eg: `127.0.0.2:6379`.

charts/istio-endpoints/templates/sidecar.yaml
```diff
@@ -0,0 +1,62 @@
+{{- if .Values.sidecar.enabled }}
+apiVersion: networking.istio.io/v1beta1
+kind: Sidecar
```
sabw8217, line 3: the expectation is that in staging/prod this would be disabled, because this should be injected by the Istio controller, right? This is for local dev where you don't have the operator/controller installed?

sabw8217: actually that looks like it's not true? I guess I don't understand how this sidecar interacts with the sidecar that's already being created by istio

PR #34, opened 2021-06-25
[simple-app] Introduce the topologySpreadConstraints setting to the PodSpec

diranged:


diranged:
@sabw8217 @ericayin This will allow us to spread our apps out across hosts and zones..

PR #35, opened 2021-08-12
[local-kafka] Add external listener

ericayin:
I wanted to be able to access `local-kafka` from services running directly on my local machine, and was able to get something working by adding an external listener (https://strimzi.io/blog/2019/04/23/accessing-kafka-part-2) that advertises `127.0.0.1:32000` for the single broker, and then exposing that address in the kind config. 

Output from `make template`:
```
# Source: local-kafka/templates/kafka.yaml
# https://github.com/strimzi/strimzi-kafka-operator/blob/master/examples/kafka/kafka-ephemeral.yaml
apiVersion: kafka.strimzi.io/v1beta2
kind: Kafka
metadata:
  name: default
spec:
  kafka:
    replicas: 1
    livenessProbe:
      initialDelaySeconds: 60
      failureThreshold: 30  # 30 * 10s
    readinessProbe:
      initialDelaySeconds: 60
      failureThreshold: 30  # 30 * 10s
    authorization:
      type: simple
      superUsers:
        - ANONYMOUS
    jvmOptions:
      '-Xmx': '384M'
      '-Xms': '128M'
    listeners:
      - name: tls
        port: 9092
        type: internal
        tls: true
        authentication:
          type: tls
      - name: plain
        port: 9093
        type: internal
        tls: false
      - configuration:
          brokers:
          - advertisedHost: 127.0.0.1
            broker: 0
            nodePort: 32000
        name: external
        port: 9094
        tls: false
        type: nodeport
    config:
      offsets.topic.replication.factor: 1
      transaction.state.log.replication.factor: 1
      transaction.state.log.min.isr: 1
      log.message.format.version: "2.8"
      inter.broker.protocol.version: "2.8"
    storage:
      type: jbod
      volumes:
      - id: 0
        type: persistent-claim
        size: 500Mi
        deleteClaim: false
  zookeeper:
    replicas: 1
    storage:
      type: persistent-claim
      size: 500Mi
      deleteClaim: false
  entityOperator:
    topicOperator: {}
    userOperator: {}
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,7 +46,7 @@ jobs:
         run: ct --config ct.yaml lint
 
       - name: Create kind cluster
-        uses: helm/kind-action@v1.0.0
+        uses: helm/kind-action@v1.2.0
         if: steps.list-changed.outputs.changed == 'true'
 
       - name: Install CRDs
--- a/charts/local-kafka/templates/kafka.yaml
+++ b/charts/local-kafka/templates/kafka.yaml
@@ -17,6 +17,8 @@ spec:
       failureThreshold: 30  # 30 * 10s
     authorization:
       type: simple
+      superUsers:
+        - ANONYMOUS
     jvmOptions:
       '-Xmx': '384M'
       '-Xms': '128M'
@@ -31,6 +33,15 @@ spec:
         port: 9093
         type: internal
         tls: false
+      - name: external
+        port: 9094
+        type: nodeport
+        tls: false
+        configuration:
+          brokers:
+            - broker: 0
+              nodePort: 32000
+              advertisedHost: 127.0.0.1
     config:
       offsets.topic.replication.factor: 1
       transaction.state.log.replication.factor: 1
--- a/charts/local-kafka/templates/kafka.yaml
+++ b/charts/local-kafka/templates/kafka.yaml
@@ -33,15 +33,7 @@ spec:
         port: 9093
         type: internal
         tls: false
-      - name: external
-        port: 9094
-        type: nodeport
-        tls: false
-        configuration:
-          brokers:
-            - broker: 0
-              nodePort: 32000
-              advertisedHost: 127.0.0.1
+      {{- toYaml .Values.listeners | nindent 6 }}
     config:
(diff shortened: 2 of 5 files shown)


charts/local-kafka/templates/kafka.yaml
```diff
@@ -17,6 +17,8 @@ spec:
       failureThreshold: 30  # 30 * 10s
     authorization:
       type: simple
+      superUsers:
+        - ANONYMOUS
```
ericayin, line 21: ran into an issue (`User:ANONYMOUS is Denied Operation`) where without tls, the user is `ANONYMOUS`, which isn't allowed to do anything currently. not sure if there's a better way to give it permissions though?

sabw8217: I think this is fine, it's local.

.github/workflows/ci.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
         run: ct --config ct.yaml lint
 
       - name: Create kind cluster
-        uses: helm/kind-action@v1.0.0
+        uses: helm/kind-action@v1.2.0
```
ericayin, line 49: https://github.com/helm/kind-action/issues/42

charts/local-kafka/templates/kafka.yaml
```diff
@@ -31,6 +33,15 @@ spec:
         port: 9093
         type: internal
         tls: false
+      - name: external
+        port: 9094
+        type: nodeport
+        tls: false
+        configuration:
```
diranged, line 40: My only suggestion would be that this listener should be probably configurable in the values file... so that if for some reason `NodePort` is blocked, they don't have to create it. Maybe the entire `listeners` key should just move into the values file?

ericayin: i moved just the external listener for now, mainly so if somebody did want to configure it, they don't have to duplicate the entire `listeners` key. `9092` and `9093` seem fairly standard for the tsl and plaintext ports, so i'm guessing that those probably won't be overwritten much? i can move the entire key if that's preferred though.

diranged (approved):

PR #36, opened 2021-08-17
[prometheus-alerts] Bump chart version to unblock release

ericayin:
https://github.com/Nextdoor/k8s-charts/runs/3353281960 is detecting [changes to the runbook](https://github.com/Nextdoor/k8s-charts/commit/87f51e566fba62abc62d584ad597b7802f82fdad), and tries to re-release the chart, but `prometheus-alerts-0.1.4` already exists. 

I wonder if the `local-kafka` chart version needs to be incremented also, since https://github.com/Nextdoor/k8s-charts/releases/tag/local-kafka-0.1.0 got created?

### DIFF
--- a/charts/prometheus-alerts/Chart.yaml
+++ b/charts/prometheus-alerts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: prometheus-alerts
 description: Helm Chart that provisions a series of common Prometheus Alerts
 type: application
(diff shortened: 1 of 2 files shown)


sabw8217 (approved):

sabw8217:
@ericayin as long as we didn't touch any of the files in that chart directory without bumping the version, I think it should be OK?

PR #43, opened 2021-08-27
[simple-app] Make the Istio Gateways more configurable

diranged:


charts/simple-app/values.yaml
```diff
@@ -243,12 +243,15 @@ virtualService:
   # https://istio.io/latest/docs/reference/config/networking/virtual-service/#VirtualService
   # for more details.
   hosts:
-    - '{{ include "simple-app.fullname" . }}.{{ .Release.Namespace }}'
-
-  # -- The name of the Istio `IngressGateway` object that this `VirtualService`
-  # will register with. The default here is a private internal gateway that is
-  # not internet-facing.
-  gateway: default-gateway
+    - '{{ include "simple-app.fullname" . }}'
```
mrterry, line 246: why did you remove the Release.Namesapce

charts/simple-app/values.yaml
```diff
@@ -243,12 +243,15 @@ virtualService:
...
+
+  # -- The name of the Istio `Gateway` resource that this `VirtualService`
+  # will register with. You can get a list of the avaialable `Gateways` by
+  # running `kubectl -n istio-system get gateways`. Not specifying a Gateway
+  # means that you are creating a VirtualService routing definition only inside
+  # of the Kubernetes cluster, which is totally reasonable if you want to do
+  # that.
+  gateways: []
```
mrterry, line 254: since FS is inside the mesh, i don't think I need to specify a gateway, but i'm not sure

PR #52, opened 2021-09-10
[simple-app] Drop the preStop default from 30s to 10s

diranged:


charts/simple-app/values.yaml
```diff
@@ -188,14 +188,18 @@ progressDeadlineSeconds: null
...
 # any inflight requests that your app is processing will fail.
+#
+# Make sure you set this to SHORTER than the terminationGracePeriod (30s
+# default) setting.
+#
 # https://docs.flagger.app/tutorials/zero-downtime-deployments#graceful-shutdown
 preStopCommand:
   - /bin/sleep
-  - '30'
+  - '10'
```
sabw8217, line 202: I think it might be worth commenting that the reason we are doing this is to give the pod time to get deregistered from the service.

sabw8217: I guess the link to the blog post is sufficient

sabw8217 (approved):

PR #59, opened 2021-09-22
[flink-job] Grant oncall engineers patch on configmaps in namespaces where the flink-job chart has been installed.

sabw8217:
…


Zedive (approved):
Looks good!

charts/flink-job/templates/operator-rolebinding.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+# This role binding allows oncall engineers to patch configmaps.
+kind: RoleBinding
+metadata:
+  name: {{ .Release.Name }}-operator-role-binding
+  namespace: {{ .Release.Namespace }}
+subjects:
+- kind: Group
```
diranged, line 8: Let's parameterize this and leave the subjects out of this codebase entirely:

```yaml
# values.yaml
operatorGroups: [system:masters]
```


```yaml
subjects:
{{- range $group := .Values.operatorGroups }}
- kind: Group
   name: {{ $group }}
   namespace: {{ .Release.Namespace }}
{{- end }}
```

This default uses a standard core K8S group (system:masters), but then lets us override it at the chart leve.

charts/flink-job/templates/operator-role.yaml
```diff
@@ -0,0 +1,13 @@
...
+# the flink operator uses to persist state about the flink cluster.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Release.Name }}-operator
+  namespace: {{ .Release.Namespace }}
+rules:
+- apiGroups: [""] # "" indicates the core API group
+  resources: ["configmaps"]
+  verbs: ["patch"]
```
pjthepooh, line 13: do we have delete permission?

sabw8217: do you need to delete these configmaps or just edit the values in them?

before we start just deleting these I think we should understand what the data in them is and how the flink operator is using it.....

pjthepooh: The one we want (need) to delete is the orphaned configmap that generated quickly and dynamically when the app was bouncing. When we force stop / force sync (to either simulate "redeploy" operation or simply want a clean state)  the app, sometimes an orphaned configmap will be left there. We have seen this might cause weird behavior in the next run of the job. So if we can delete them, it would give us a clean state to start with.

pjthepooh: Screen Shot 2021-09-21 at 5 33 49 PM
one of the `interaction-store-<id>-ledger` should be removed. And the dispatcher ledger can be patched or removed.

PR #62, opened 2021-09-23
[flink-job] Fix the role/rolebindings and introduce metric relabeling configs

diranged:
**What did I want?**

First, I noticed that the `RoleBinding` yaml was broken, which meant that the chart wouldn't work. Second, I wanted to be able to drop certain high cardinality and low-value labels from the metrics before they are scraped by Prometheus. The `task_attempt_id` label is the single most expensive label (by 4x the next most) in our entire prometheus system at the moment.


charts/flink-job/templates/operator-rolebinding.yaml
```diff
@@ -7,8 +7,8 @@ metadata:
 subjects:
   {{- range $group := .Values.operatorGroups }}
   - kind: Group
-     name: {{ $group }}
-     namespace: {{ .Release.Namespace }}
+    name: {{ $group }}
+    namespace: {{ $.Release.Namespace }}
```
sabw8217, line 11: we need to split the helm lint and helm test stuff apart...helm test has never worked for this chart and it's masking failures that I think would be caught by helm lint

sabw8217 (approved):

pjthepooh (approved):

Zedive (approved):
🙌

PR #64, opened 2021-09-24
[flink-job] Expose SampleLimit in flink-job podMonitor

pjthepooh:
expose smaplelimit in flink-job podmonitor


charts/flink-job/templates/podmonitor.yaml
```diff
@@ -27,4 +27,7 @@ spec:
       relabelings:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+  {{- with .Values.podMonitor.SampleLimit }}
+  SampleLimit: {{ . }}
```
sabw8217, line 31: because you aren't doing anything with this, it would be a little more idiomatic to just have `SampleLimit: {{ .Values.podMonitor.sampleLimit`.

I would also put an `{{- if .Values.podMonitor.sampleLimit }} .. {{- end}}` around this, so you don't even set this to anything (meaning it would use the global default) if a value is not set in a values file.

pjthepooh: for my education, how is `{{- if .Values.podMonitor.sampleLimit }}` diff from `{{- with .Values.podMonitor.SampleLimit }}` and use `.` for value

charts/flink-job/templates/podmonitor.yaml
```diff
@@ -27,4 +27,7 @@ spec:
       relabelings:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+  {{- if .Values.podMonitor.sampleLimit }}
```
diranged, line 30: Use the `with` syntax whenever you can - it's cleaner and less verbose.

charts/flink-job/templates/podmonitor.yaml
```diff
@@ -27,4 +27,7 @@ spec:
...
+  SampleLimit: {{ .Values.podMonitor.sampleLimit }}
```
diranged, line 31: Pretty sure this needs to be `sampleLimit`

charts/flink-job/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "1.0"
 description: Flink job cluster on k8s
 name: flink-job
-version: 0.1.1
+version: 0.1.2
```
diranged, line 5: Always set a default in the `values.yaml` file and add an appropriate documentation string.

pjthepooh: sure, i just couldn't find the default value so didn't want to cause issue when others update the chart.

charts/flink-job/templates/podmonitor.yaml
pjthepooh: I used `with` but was suggested to use `if`. cc: @sabw8217

charts/flink-job/templates/podmonitor.yaml
sabw8217: `with` defines a scope: https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with

`if` is a conditional block that does not modify scope: https://helm.sh/docs/chart_template_guide/control_structures/#ifelse

sabw8217 (approved):

PR #65, opened 2021-09-25
[flink-job] Add RocksDB PVC

pjthepooh:
Adds a new `PersistentVolumeClaim` dedicated for RocksDB with default disabled using EBS

### DIFF
--- a/charts/flink-job/README.md
+++ b/charts/flink-job/README.md
@@ -2,7 +2,7 @@
 
 Flink job cluster on k8s
 
-![Version: 0.1.2](https://img.shields.io/badge/Version-0.1.2-informational?style=flat-square) ![AppVersion: 1.0](https://img.shields.io/badge/AppVersion-1.0-informational?style=flat-square)
+![Version: 0.1.3](https://img.shields.io/badge/Version-0.1.3-informational?style=flat-square) ![AppVersion: 1.0](https://img.shields.io/badge/AppVersion-1.0-informational?style=flat-square)
 
 This chart deploys a flink job cluster and runs a simple word counting flink app as an example.
 This chart includes some production ready set-ups such as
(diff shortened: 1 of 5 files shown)


charts/flink-job/README.md
```diff
@@ -51,6 +51,11 @@ See metrics reporter in the flink properties for more details.
 | podMonitor.sampleLimit | int | `2000` | (`int`) Per-scrape limit on number of scraped samples that will be accepted. |
 | podMonitor.scrapeInterval | string | `"15s"` | (`string`) The frequency in which to scrape metrics. |
 | pvc | object | `{"storage":"1Gi","storageClassName":"efs"}` | Configuration of the PersistentVolume for storing savepoints. |
+| pvcRocksdb | object | `{"enabled":false,"rocksdbDir":"/rocksdb","storage":"10Gi","storageClassName":"gp3-encrypted"}` | Configuration of rocksdb PVC |
+| pvcRocksdb.enabled | bool | `false` | (`boolean`) whether to enable RocksDB PVC |
+| pvcRocksdb.rocksdbDir | string | `"/rocksdb"` | (`string`) The mount path of the rocksdb volume |
+| pvcRocksdb.storage | string | `"10Gi"` | (`string`) The size of the storage |
+| pvcRocksdb.storageClassName | string | `"gp3-encrypted"` | (`string`) The class name of the storage. The list of available storage classes is from `k get storageclasses` |
```
diranged, line 58: 1. Remove the default storage-class name here ... its internal to us, and is not a relevant default for anyone else.
2. You don't need a storageClassName here actually at all - if you don't pick one, the cluster uses its default. You _can_ put the option in, but if so, the general pattern is to not set it to anything explicit, and instead make sure the helm chart user sets it to something that is relevant to them.

charts/flink-job/README.md
```diff
@@ -51,6 +51,11 @@ See metrics reporter in the flink properties for more details.
 | podMonitor.sampleLimit | int | `2000` | (`int`) Per-scrape limit on number of scraped samples that will be accepted. |
 | podMonitor.scrapeInterval | string | `"15s"` | (`string`) The frequency in which to scrape metrics. |
 | pvc | object | `{"storage":"1Gi","storageClassName":"efs"}` | Configuration of the PersistentVolume for storing savepoints. |
+| pvcRocksdb | object | `{"enabled":false,"rocksdbDir":"/rocksdb","storage":"10Gi","storageClassName":"gp3-encrypted"}` | Configuration of rocksdb PVC |
```
diranged, line 54: Personally I'd have a key named `rocksdb` and inside it I would have the options like this:

```yaml
rocksdb:
  enabled: true|false  # (this would somehow configure the flink job itself to use rocksdb?)
  pvc:
    enabled: true|false
    storageClassName: null  # if set, it would then set it in the template
    storage: 10Gi
    path: /rocksdb
```

charts/flink-job/templates/rocksdb-pvc.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.pvcRocksdb.enabled -}}
```
diranged, line 1: So - in this case, I don't think this is right. You're not creating a single PVC, but you actually want to create a PVC for each and every Pod that runs in the StatefulSet that the Flink Operator is managing. Your going to want something like https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#components - see the `volumeClaimTemplate` setting. There is a discussion about this exact scenario at https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/issues/353.

@sabw8217 When I read https://github.com/shashken/flink-on-k8s-operator/blob/master/docs/crd_v1alpha1.md, it is not immediately clear to me that they support setting up the `volumeClaimTemplate` setting. Am I missing something obvious?

sabw8217: That's not the CRD we are using.

https://github.com/GoogleCloudPlatform/flink-on-k8s-operator/blob/master/docs/crd.md

sabw8217: but @diranged is correct, you do not need this. The PVC is generated for you from the template in the StatefulSet, you do not need to create it separately.

diranged: 🤦  thanks aaron.. oops..that CRD does have the volumeClaimTemplates section..

pjthepooh:
Thanks for the review! This appears to be not needed as the CRD we use already has `taskManager.volumeClaimTemplates` and we can fill it in app level when needed. Closing this.

PR #67, opened 2021-10-01
[stateful-app] Create a new Stateful App chart similar to Simple-App

diranged:


masterginger (approved):
Currently the localstack pod in nd-hybrid uses statefulset, the change about service and statefulset in this PR look good to me.

PR #68, opened 2021-10-02
[flink-job] Expose flinkVersion field

pjthepooh:
Adding this field for Spotify flink operator we recently switched over to in https://github.com/Nextdoor/k8s/pull/508

### DIFF
--- a/charts/flink-job/Chart.yaml
+++ b/charts/flink-job/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "1.0"
 description: Flink job cluster on k8s
 name: flink-job
-version: 0.1.2
+version: 0.1.3
 maintainers:
   - name: Zedive
     email: albert@nextdoor.com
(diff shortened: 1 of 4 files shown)


sabw8217 (approved):
we confirmed there aren't any other setting we need to add?

diranged (changes requested):
Please don’t set the default value to 1.13.1 … instead, leave the value string empty in the values.yaml file.

pjthepooh:
> we confirmed there aren't any other setting we need to add?

cpu limit as well, but can be handled in app level

We may want to expose `Status` field too (i need to read up how to use it), but lets do it in another PR and later.

diranged:
> > we confirmed there aren't any other setting we need to add?
> 
> cpu limit as well, but can be handled in app level
> 
> We may want to expose `Status` field too (i need to read up how to use it), but lets do it in another PR and later.

The status field is read-only  - its filled in by the operator, not by us.. so nothing to do here in the chart for that.

diranged (approved):

PR #69, opened 2021-10-04
Add an alarm on 5xx to the simple-app chart.

sabw8217:


charts/simple-app/values.yaml
```diff
@@ -457,3 +457,21 @@ tests:
...
       # If this is left empty, the default is "latest"
       tag:
+
+# -- The URL of the runbook for this service.
+runbookUrl:
+
+alerts:
+  severity: critical
+
+virtualService5xxMonitor:
```
diranged, line 467: I would move this to the `virtualService` key:

```yaml
# values.yaml
virtualService:
  ....
  serviceMonitor:
    enabled: true
    requestRules:
      enabled: true
      period: 5m
      threshold: 0.00....
```

diranged: this would make it more similar to https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/values.yaml#L85-L96, and also keeps the serviceMonitor config nested within the virtualservice

PR #71, opened 2021-10-06
Remove chomps on toYAML.

sabw8217:


charts/flink-job/templates/flinkjobcluster.yaml
```diff
@@ -36,7 +36,7 @@ spec:
           claimName: {{ template "flink-job-cluster.fullname" . }}-savepoint
   {{- end }}
   {{- if .Values.jobManager.volumes }}
-    {{- toYaml .Values.jobManager.volumes | indent 6}}
+    {{ toYaml .Values.jobManager.volumes | indent 6}}
```
sabw8217, line 39: I think the key here is:
1. no indenting on something where you are going to use the `indent` filter
2. no "chomps" on `toYaml` if it is part of a list

I'm not totally sure this works if savepoints are not enabled though, will test locally

Zedive: I see. I can test that out manually. But agree having a unit test of some sort would be nicer.

Zedive: The fix should work 

```
    volumes:
      - name: savepoint-storage
        persistentVolumeClaim:
          claimName: word-counting-cluster-savepoint
      - configMap:
          name: event-proxy-properties
        name: config
    volumeMounts:
      - name: savepoint-storage
        mountPath: /savepoint
      - mountPath: /config
        name: config
        readOnly: true
```

PR #75, opened 2021-10-12
[flink-job] Fix indentation and reformat flinkjobcluster yaml

pjthepooh:
fix indentation in flink-job

### DIFF
--- a/charts/flink-job/templates/flinkjobcluster.yaml
+++ b/charts/flink-job/templates/flinkjobcluster.yaml
@@ -13,95 +13,95 @@ spec:
     pullPolicy: IfNotPresent
   batchSchedulerName: {{ .Values.batchSchedulerName }}
   envVars:
-{{ toYaml .Values.envVars | indent 4 }}
+    {{- toYaml .Values.envVars | nindent 4 }}
   serviceAccountName: {{ template "flink-job-cluster.fullname" . }}
   jobManager:
     accessScope: {{ .Values.jobManager.accessScope }}
     ports:
       ui: {{ .Values.jobManager.ports.ui }}
       # TODO (Aaron/Albert) expose the other ports?
-{{- if .Values.jobManager.metrics.enabled }}
+    {{- if .Values.jobManager.metrics.enabled }}
     extraPorts:
-{{ toYaml .Values.jobManager.metrics.extraPorts | indent 6 }}
-{{- end }}
+      {{- toYaml .Values.jobManager.metrics.extraPorts | nindent 6 }}
+    {{- end }}
     # TODO (Aaron/Albert) expose ingress?
     resources:
-{{ toYaml .Values.jobManager.resources | indent 6 }}
+      {{- toYaml .Values.jobManager.resources | nindent 6 }}
     # TODO (Aaron/Albert) expose memory configs?
-{{- if or (.Values.jobManager.volumes) (.Values.savepoints.enabled) }}
+    {{- if or (.Values.jobManager.volumes) (.Values.savepoints.enabled) }}
     volumes:
-  {{- if .Values.savepoints.enabled }}
+      {{- if .Values.savepoints.enabled }}
       - name: savepoint-storage
         persistentVolumeClaim:
           claimName: {{ template "flink-job-cluster.fullname" . }}-savepoint
-  {{- end }}
-  {{- with .Values.jobManager.volumes }}
-  {{- toYaml . | nindent 6 }}
-  {{- end }}
-{{- end }}
-{{- if or (.Values.jobManager.volumeMounts) (.Values.savepoints.enabled) }}
+      {{- end }}
+      {{- with .Values.jobManager.volumes }}
+      {{- toYaml . | nindent 6 }}
+      {{- end }}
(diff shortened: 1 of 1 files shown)


pjthepooh:
@sabw8217 not sure why the test is failing

charts/flink-job/templates/flinkjobcluster.yaml
```diff
@@ -36,7 +36,7 @@ spec:
           claimName: {{ template "flink-job-cluster.fullname" . }}-savepoint
   {{- end }}
   {{- if .Values.jobManager.volumes }}
-    {{ toYaml .Values.jobManager.volumes | indent 6}}
+{{ toYaml .Values.jobManager.volumes | indent 6}}
```
diranged, line 39: You actually _want_ this indented, and you want to use `| nindent` instead of `|indent`

charts/flink-job/templates/flinkjobcluster.yaml
```diff
@@ -46,7 +46,7 @@ spec:
         mountPath: {{ .Values.savepoints.savepointDir }}
   {{- end }}
   {{- if .Values.jobManager.volumeMounts }}
-    {{- toYaml .Values.jobManager.volumeMounts | indent 6}}
+{{ toYaml .Values.jobManager.volumeMounts | indent 6}}
```
diranged, line 49: same as above..

charts/flink-job/templates/flinkjobcluster.yaml
diranged: The right way to do this is with nindent.. almost always.

pjthepooh: sounds good. updated!

charts/flink-job/templates/flinkjobcluster.yaml
```diff
@@ -36,7 +36,7 @@ spec:
           claimName: {{ template "flink-job-cluster.fullname" . }}-savepoint
   {{- end }}
   {{- if .Values.jobManager.volumes }}
-    {{ toYaml .Values.jobManager.volumes | indent 6}}
+    {{- toYaml .Values.jobManager.volumes | nindent 6}}
```
diranged, line 39: Hard to see in this PR - but you have 4 spaces here in the beginning, and you are indenting by 6.. i think you want those to match. So you either want it to be 4 or 6 depending on the format.

pjthepooh: Rendered locally here what I got, 
with nindent 6
```
    volumes:
      - hostPath:
          path: /temp/savepoints
          type: DirectoryOrCreate
        name: local-savepoint-dir
    volumeMounts:
      - mountPath: /temp/savepoints
        name: local-savepoint-dir
```
with nindent 4
```
    volumes:
    - hostPath:
        path: /temp/savepoints
        type: DirectoryOrCreate
      name: local-savepoint-dir
    volumeMounts:
    - mountPath: /temp/savepoints
      name: local-savepoint-dir
```
it seems effect is the same, but format is diff

pjthepooh: looks `nindent 6` is consistent better with other indentation in place, as it should be one level under `volumes` that has 4 spaces

diranged: The number of spaces you are indenting is specific to the key in the Yaml that you are trying to insert data into. The key you're inserting into is at https://github.com/Nextdoor/k8s-charts/blob/f9f87badf9d099a7c9178cc1b6ec6366f697d665/charts/flink-job/templates/flinkjobcluster.yaml#L43 and has 4 spaces in front of it. Because it's a YAML Array, you can EITHER indent it 4 or 6 spaces, but you cannot mix and match. The "savepoint" blob (https://github.com/Nextdoor/k8s-charts/blob/f9f87badf9d099a7c9178cc1b6ec6366f697d665/charts/flink-job/templates/flinkjobcluster.yaml#L44-L47) is indenting 6 spaces it looks like.

So the current code looks like this:
```yaml
{{- if or (.Values.jobManager.volumeMounts) (.Values.savepoints.enabled) }}
    volumeMounts:
  {{- if .Values.savepoints.enabled }}
      - name: savepoint-storage
        mountPath: {{ .Values.savepoints.savepointDir }}
  {{- end }}
  {{- if .Values.jobManager.volumeMounts }}
    {{- toYaml .Values.jobManager.volumeMounts | nindent 6}}
  {{- end }}
{{- end }}
```

I would reformat it ... the fact that some of the conditions are "less indented" than the text they then create is really confusing. Instead, I'd do this:

```yaml
    {{- if or (.Values.jobManager.volumeMounts) (.Values.savepoints.enabled) }}
    volumeMounts:
      {{- if .Values.savepoints.enabled }}
      - name: savepoint-storage
        mountPath: {{ .Values.savepoints.savepointDir }}
      {{- end }}
      {{- with .Values.jobManager.volumeMounts }}
      {{- toYaml . | nindent 6}}
      {{- end }}
    {{- end }}
```

By indenting everything to the proper level, it's easier to understand what the indent levels are supposed to look like.

PR #76, opened 2021-10-15
[flink-job] Update yaml to be compatible with updated flink operator

pjthepooh:
To be compatible with the change on flink operator https://github.com/Nextdoor/k8s/pull/565.

### DIFF
--- a/charts/flink-job/README.md
+++ b/charts/flink-job/README.md
@@ -2,7 +2,7 @@
 
 Flink job cluster on k8s
 
-![Version: 0.1.7](https://img.shields.io/badge/Version-0.1.7-informational?style=flat-square) ![AppVersion: 1.0](https://img.shields.io/badge/AppVersion-1.0-informational?style=flat-square)
+![Version: 0.1.8](https://img.shields.io/badge/Version-0.1.8-informational?style=flat-square) ![AppVersion: 1.0](https://img.shields.io/badge/AppVersion-1.0-informational?style=flat-square)
 
 This chart deploys a flink job cluster and runs a simple word counting flink app as an example.
 This chart includes some production ready set-ups such as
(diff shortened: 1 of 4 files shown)


charts/flink-job/README.md
```diff
@@ -42,10 +42,14 @@ See metrics reporter in the flink properties for more details.
 | job.restartPolicy | string | `"FromSavepointOnFailure"` | (String) Restart policy when the job fails, enum("Never", "FromSavepointOnFailure") |
 | job.savepointGeneration | `int` | `nil` | Update this field to jobStatus.savepointGeneration + 1 for a running job cluster to trigger a new savepoint to savepointsDir on demand. |
 | job.savepointsDir | string | `"/savepoint"` | (String) Directory to store automatically taken savepoints |
-| job.takeSavepointOnUpgrade | bool | `true` | (bool) Should take savepoint before upgrading the job |
+| job.takeSavepointOnUpdate | bool | `true` | (bool) Should take savepoint before upgrading the job |
 | jobManager.accessScope | string | `"Cluster"` | (String) Access scope of the JobManager service. enum("Cluster", "VPC", "External", "NodePort", "Headless") |
+| jobManager.memoryProcessRatio | int | `80` | ('list') Percentage of memory process, as a safety margin to avoid OOM kill |
```
pjthepooh, line 47: These fields need to be exposed otherwise will get null point exceptions.

charts/flink-job/values.yaml
```diff
@@ -122,6 +141,8 @@ taskManager:
   volumes: []
   # -- ('list') volumeMounts for the TaskManager containers
   volumeMounts: []
+  # -- ('list') Percentage of memory process, as a safety margin to avoid OOM kill
```
Zedive, line 144: this one is `int`

pjthepooh: good catch

PR #83, opened 2021-10-27
Move 5xx alarm from simple-app to prometheus-alerts.

vsood5:
Adding alarm for highLatency of service requests.


RKTMN:
There's no way this is getting merged in at this point...

PR #84, opened 2021-10-27
[flink-operator] Create flink-operator chart

pjthepooh:
Add a helm chart for flink operator without CRD


charts/flink-operator/templates/misc.yaml
```diff
@@ -0,0 +1,39 @@
...
+      control-plane: controller-manager
+  policyTypes: [Ingress]
+  ingress:
+    - ports:
+      - port: 9443
+        protocol: TCP
+---
+# Grant the Flink Operator carte-blanche to connect to and manage
+# JobManager pods throughout the cluster.
+{{ if .Values.enableGlobalNetworkPolicy }}
```
sabw8217, line 21: so I think we can get rid of this now, because the flink operator should be running in the same namespace as the flink job is it managing?

charts/flink-operator/templates/clusterroles.yaml
```diff
@@ -0,0 +1,26 @@
...
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+  - apiGroups: [ flinkoperator.k8s.io ]
+    resources: [ "*" ]
+    verbs: [ create, delete, deletecollection, patch, update]
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
sabw8217, line 16: I think you may be able to downgrade this to being just a `Role` instead of a `ClusterRole` as well, but I would need to check more carefully...

pjthepooh: This makes sense. Looks like the diff is just scope (namespace vs cluster), but I have verified `ClusterRole`, will verify Role as well.

charts/flink-operator/Chart.yaml
```diff
@@ -0,0 +1,19 @@
...
+version: 0.0.1
+appVersion: v0.3.0
+sources:
+  - https://github.com/spotify/flink-on-k8s-operator/releases
+maintainers:
+  - name: pjthepooh
+    email: pjhuang@nextdoor.com
+dependencies:
+  - name: flink-operator-crd
+    version: 0.3.0
```
pjthepooh, line 13: this needs to be 0.3.1?

charts/flink-operator/templates/proxy-role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: {{ if .Values.namespaceScoped }}Role{{ else }}ClusterRole{{- end }}
```
pjthepooh, line 2: @diranged for `Role`, do we need to set namespace in `metadata`?

charts/flink-operator/Chart.yaml
diranged: ah its a nice to have, but not critical.. and right now the tests would fail because this PR includes the 0.3.1 changes, but they haven't been merged yet so there is no downloadable chart yet anyways

charts/flink-operator/templates/proxy-role.yaml
diranged: No - all resources by default created by Helm charts are automatically put into a particular target namespace.. it's actually a bad practice to hard-code a `namespace` key when its not necessary.

charts/flink-operator/templates/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - v1
+  clientConfig:
+    service:
+      name: {{ .Release.Name }}-webhook-service
+      namespace: {{ .Release.Namespace }}
+      path: /mutate-flinkoperator-k8s-io-v1beta1-flinkcluster
+  failurePolicy: Fail
+  name: mflinkcluster.flinkoperator.k8s.io
+  {{- if .Values.namespaceScoped }}
+  namespaceSelector:
```
pjthepooh, line 19: @diranged I think this namespaceSelector is causing issue on the job submitter hanging in running status after submission.

pjthepooh: The way the operator deploy via the make target from the repo has this extra logic to set the label flink-operator-namespace. https://github.com/spotify/flink-on-k8s-operator/blob/5db74d5ff6562af1b3bd5fc9c70a4f4084000263/Makefile#L98-L103

charts/flink-operator/templates/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,37 @@
...
+    matchExpressions:
+      - key: flink-operator-namespace
```
diranged, line 21: After poking around, I suggest changing this to `kubernetes.io/metadata.name`.. that label is automatically applied to namespaces and it matches the actual name of the namespace:

```
$ k get ns -l kubernetes.io/metadata.name=kafka
NAME    STATUS   AGE
kafka   Active   179d
```

pjthepooh: good find! changed it

charts/flink-operator/templates/mutatingwebhookconfiguration.yaml
pjthepooh: updated hard coded key `flink-operator-namespace` to `kubernetes.io/metadata.name` instead to avoid setting the label. 

Also made change in `flilnk-job` to resolve job submitter hanging and other issues due to not using webhook.

diranged: awsome!

PR #86, opened 2021-10-29
[flink-operator-crd] Add flink-operator-crd chart

pjthepooh:
Add a helm chart only contains CRD for flink operator


sabw8217 (commented):
you should get @diranged to confirm this is how it is supposed to be...I thought we were going to use the `crds` folder and `--skip-crds` in some situations

https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

pjthepooh:
> you should get @diranged to confirm this is how it is supposed to be...I thought we were going to use the `crds` folder and `--skip-crds` in some situations
> 
> https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

I think if we separate CRD and other resources in diff charts (method 2 in the link), should be fine?

charts/flink-operator-crd/templates/flink-operator-crd.yaml
```diff
@@ -0,0 +1,6378 @@
...
+    controller-gen.kubebuilder.io/version: v0.6.2
+  name: flinkclusters.flinkoperator.k8s.io
+spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          name: flink-operator-webhook-service
+          namespace: {{ .Release.Namespace }}
```
diranged, line 19: This file does not get templatized ... remove all the references to `{{ .Release.Namespace }}`

charts/flink-operator-crd/templates/flink-operator-crd.yaml
```diff
@@ -0,0 +1,6378 @@
+# This is originally from https://github.com/spotify/flink-on-k8s-operator/releases/download/v0.3.0/flink-operator.yaml
+# with modifications for general usage
+#
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.kubernetes.io: unapproved
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/flink-operator-serving-cert
```
diranged, line 9: remove this i think

diranged:
> you should get @diranged to confirm this is how it is supposed to be...I thought we were going to use the `crds` folder and `--skip-crds` in some situations
> 
> https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

You are correct - @pjthepooh please use the `crds/` directory, not a template.

charts/flink-operator-crd/templates/flink-operator-crd.yaml
pjthepooh: @diranged should we use a specific namespace, or this field is not needed? (it was `flink-operator-system` originally)

diranged: I believe we should yank out or comment out that section.

pjthepooh: namespace is a required field, so commenting out webhook conversion section https://v1-20.docs.kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion

pjthepooh: Looks like we need this webhood service though? @diranged

pjthepooh: Verified locally with this section commented out and had no issue.

PR #88, opened 2021-11-05
[flink-job] Expose job mode and JM replicas

pjthepooh:
Flink operator uses mutating webhook to set default values for some fields in CRDs, https://github.com/spotify/flink-on-k8s-operator/blob/5f2c73a16f9040cb06cd9665e6a59b77d2280709/api/v1beta1/flinkcluster_default.go. As we try to scope down the operator to namespace level with webhook disabled, we need to expose these fields with default values set.


sabw8217 (approved):

sabw8217:
damn I thought someone had fixed those tests, I can look at it next week

PR #92, opened 2021-11-29
[istio-endpoints] Allow custom annotations on sidecar

ericayin:
I have a job that makes requests to pods in another namespace in the mesh. I want to run it as a `PreSync` hook, but that means that the custom sidecar configuration won't be applied yet at that point. I also want the sidecar to be created before the job is run, and was thinking of using [waves](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-waves/) to do that?

### DIFF
--- a/charts/istio-endpoints/templates/sidecar.yaml
+++ b/charts/istio-endpoints/templates/sidecar.yaml
@@ -6,6 +6,12 @@ metadata:
   name: {{ include "istio-endpoints.fullname" . }}-proxy-config
   labels:
     {{- include "istio-endpoints.labels" . | nindent 4 }}
+  {{- with .Values.sidecar.annotations }}
+  annotations:
+    {{- range $key, $value := . }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+  {{- end }}
 spec:
   workloadSelector:
     labels:
--- a/charts/istio-endpoints/Chart.yaml
+++ b/charts/istio-endpoints/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: istio-endpoints
 description: Per-Namespace Istio Configuration Chart
 type: application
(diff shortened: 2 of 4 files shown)


charts/istio-endpoints/templates/sidecar.yaml
```diff
@@ -6,6 +6,10 @@ metadata:
   name: {{ include "istio-endpoints.fullname" . }}-proxy-config
   labels:
     {{- include "istio-endpoints.labels" . | nindent 4 }}
+  {{- with .Values.sidecar.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
```
ericayin, line 11: if this is too much flexibility, we could add more narrowly scoped values (instead of just `annotations`) that would only add the specific annotations for the sync phases / waves?

charts/istio-endpoints/templates/sidecar.yaml
```diff
@@ -6,6 +6,12 @@ metadata:
   name: {{ include "istio-endpoints.fullname" . }}-proxy-config
   labels:
     {{- include "istio-endpoints.labels" . | nindent 4 }}
+  {{- with .Values.sidecar.annotations }}
+  annotations:
```
diranged, line 10: I don't see anything specifically wrong with this, so it's fine. I am not sure that Argo Sync Waves are exactly the right answer to the problem though.

diranged (approved):

charts/istio-endpoints/templates/sidecar.yaml
ericayin: just noting this for future reference: since kubernetes operates under the expectation of eventual consistency, trying to enforce the ordering of specific resources may cause other issues/side effects. going to try this and see if it's okay for now, or if we need a different approach.

PR #97, opened 2022-01-12
[simple-app] Import all of the relevant PrometheusRules from the prometheus-alerts chart

diranged:
The goal of this PR is to make sure that each time a `simple-app` chart
is launched, it automatically includes relevant alerts for the
application itself. Many of these alerts are tricky because we have to
use `pod=~...` regex matches to scope things correctly.

The intention is that `simple-app` here will only have alerts relevant
to its own configuration... so if a chart uses several `simple-app`
charts, we'll end up with a different set of Prometheus Rules for each
chart.

### DIFF
--- a/charts/istio-alerts/templates/service-prometheusrule.yaml
+++ b/charts/istio-alerts/templates/service-prometheusrule.yaml
@@ -1,42 +1,34 @@
-{{ $values     := .Values }}
-{{ $release    := .Release }}
-
+{{ $values           := .Values }}
+{{ $release          := .Release }}
 {{ if .Values.serviceRules.enabled }}
+
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
-  name: {{ .Release.Name }}-{{ lower .Chart.Name }}-service-rules
+  name: {{ .Release.Name }}-service-rules
   annotations:
     nextdoor.com/chart: {{ .Values.chart_name }}
     nextdoor.com/source: {{ .Values.chart_source }}
 spec:
   groups:
-  - name: {{ .Release.Namespace }}.{{ .Release.Name }}.{{ .Chart.Name }}.IstioServiceRules
+  - name: {{ .Release.Name }}.{{ .Release.Namespace }}.serviceRules
     rules:
     {{ with .Values.serviceRules.http5XXMonitor }}
     {{ if .enabled }}
-    - alert: 5xx-Rate-Too-High
+    - alert: {{ $release.Name }}-5xx-Rate-Too-High
       annotations:
         summary: {{`High rate of 5xx errors from {{$labels.destination_service_name}} in namespace {{$labels.namespace}}`}}
         runbook_url: {{ $values.defaults.runbookUrl }}#5xx-Rate-Too-High
         description: >-
           High rate of 5xx responses from the {{`$labels.destination_service_name`}} VirtualService in namespace {{`{{$labels.namespace}}`}}.
       expr: >-
-        sum by (destination_service_name) (
-            increase(istio_requests_total{response_code=~"5.*", destination_service_namespace="{{ $release.Namespace }}"}[60s])
-        )
-
-          /
-
-        sum by (destination_service_name) (
-            increase(istio_requests_total{destination_service_namespace="{{ $release.Namespace }}"}[60s])
-        )
-          > {{ .threshold }}
+        sum by (destination_service_name) (increase(istio_requests_total{response_code=~"5.*", destination_service_namespace="{{ $release.Namespace }}"}[60s]))
+         / sum by (destination_service_name) (increase(istio_requests_total{destination_service_namespace="{{ $release.Namespace }}"}[60s])) > {{ .threshold }}
       for: {{ .for }}
       labels:
         severity: {{ .severity }}
-        {{- with $values.defaults.additionalRuleLabels}}
-        {{ toYaml . | nindent 8}}
+        {{- if $values.defaults.additionalRuleLabels }}
+        {{ toYaml $values.defaults.additionalRuleLabels | nindent 8 }}
         {{- end }}
     {{- end }}
     {{- end }}
@@ -44,7 +36,7 @@ spec:
      # Adding latency alarm to trigger when request latency is above > threshold
     {{ with .Values.serviceRules.highRequestLatency }}
     {{ if .enabled }}
-    - alert: HighRequestLatency
+    - alert: {{ $release.Name }}-HighRequestLatency
       annotations:
         summary: Processes experience latency above {{ .threshold }}.
         runbook_url: {{ $values.defaults.runbookUrl}} #HighRequestLatency
@@ -53,17 +45,13 @@ spec:
           namespace {{ $labels.namespace }} for container {{ $labels.container
           }} in pod {{ $labels.pod }}.`}}
       expr: |-
-        sum by (destination_service_name) (
-          irate(
-            istio_request_duration_milliseconds_bucket{reporter=~"destination",destination_service_namespace="{{ $release.Namespace }}"}[1m]
-          ) / 1000
-        )
-          > {{ .threshold }}
+        sum by (destination_service_name) (irate(istio_request_duration_milliseconds_bucket{reporter=~"destination",destination_service_namespace="{{ $release.Namespace }}"}[1m]) / 1000)
+        > {{ .threshold }}
       for: {{ .for }}
       labels:
         severity: {{ .severity }}
-        {{- with $values.defaults.additionalRuleLabels}}
-        {{ toYaml . | nindent 8}}
+        {{- if $values.defaults.additionalRuleLabels}}
+        {{ toYaml $values.defaults.additionalRuleLabels | nindent 8}}
         {{- end }}
     {{- end }}
   {{- end }}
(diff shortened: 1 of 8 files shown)


charts/simple-app/templates/prometheusrules.yaml
```diff
@@ -0,0 +1,265 @@
...
+          Pod {{`{{$labels.pod}}`}} in namespace {{`{{$labels.namespace}}`}}
+          has a container that has been terminated ({{`{{ $value }}`}} times) due to
+          {{`{{$labels.reason}}`}} in the last {{ .for }}.
+      expr: |-
+        sum by (container, namespace, pod, reason) (
+          sum_over_time(
+            kube_pod_container_status_terminated_reason{
+              reason=~"{{ join "|" .reasons }}",
+              namespace="{{ $targetNamespace }}",
+              pod=~"{{ include "simple-app.fullname" $global }}-.*"
```
sabw8217, line 33: if we are confident this pattern matching will always work, it seems a little sus to me

charts/simple-app/templates/prometheusrules.yaml
```diff
@@ -0,0 +1,265 @@
...
+      annotations:
+        summary: Deployment generation mismatch due to possible roll-back
+        runbook_url: {{ $runbookUrl }}#deploymentgenerationmismatch
+        description: >-
+          Deployment generation for {{`{{`}} $labels.namespace
+          {{`}}`}}/{{`{{`}} $labels.deployment {{`}}`}} does not match, this
+          indicates that the Deployment has failed but has not been rolled
+          back.
+      expr: |-
+        kube_deployment_status_observed_generation{job="kube-state-metrics", namespace=~"{{ $targetNamespace }}"}
```
sabw8217, line 171: should have a filter on `deployment="{{ include "simple-app.fullname" . }}"`

charts/simple-app/templates/prometheusrules.yaml
```diff
@@ -0,0 +1,265 @@
...
+          Deployment {{`{{`}} $labels.namespace {{`}}`}}/{{`{{`}}
+          $labels.deployment {{`}}`}} has not matched the expected number of
+          replicas for longer than {{ .for }}.
+      expr: |-
+        (
+          kube_deployment_spec_replicas{job="kube-state-metrics", namespace=~"{{ $targetNamespace }}"}
+            !=
+          kube_deployment_status_replicas_available{job="kube-state-metrics", namespace=~"{{ $targetNamespace }}"}
+        ) and (
+          changes(kube_deployment_status_replicas_updated{job="kube-state-metrics", namespace=~"{{ $targetNamespace }}"}[5m])
```
sabw8217, line 197: ditto

charts/simple-app/templates/prometheusrules.yaml
```diff
@@ -0,0 +1,265 @@
...
+        {{- with $values.prometheusRules.additionalRuleLabels }}
+        {{ toYaml . | nindent 8 }}
+        {{- end }}
+    {{- end }}
+
+  {{- if .Values.autoscaling.enabled }}
+  - name: {{ .Release.Namespace }}.{{ .Release.Name }}.{{ .Chart.Name }}.HorizontalPodAutoscalerRules
+    rules:
+
+    {{ with .Values.prometheusRules.HpaReplicasMismatch -}}
```
sabw8217, line 213: maybe should have `and .Values.autoscaling.enabled`?

charts/simple-app/templates/prometheusrules.yaml
```diff
@@ -0,0 +1,265 @@
...
+        changes(kube_horizontalpodautoscaler_status_current_replicas[15m]) == 0
+      for: {{ .for }}
+      labels:
+        severity: {{ .severity }}
+        {{- with $values.prometheusRules.additionalRuleLabels }}
+        {{ toYaml . | nindent 8 }}
+        {{- end }}
+    {{- end }}
+
+    {{ with .Values.prometheusRules.HpaMaxedOut -}}
```
sabw8217, line 244: ditto

charts/simple-app/values.yaml
```diff
@@ -482,3 +482,69 @@ runbookUrl:
...
+  # -- (`bool`) Whether or not to enable the container rules template
+  enabled: true
+
+  # -- (`map`) Additional custom labels attached to every PrometheusRule
+  additionalRuleLabels: {}
+
+  # -- Monitors Pods for Containers that are terminated either for unexpected
+  # reasons like ContainerCannotRun. If that number breaches the $threshold (1)
+  # for $for (1m), then it will alert.
+  PodContainerTerminated:
```
sabw8217, line 501: these should probably include an `enabled` flag to allow for granularly turning them off

charts/simple-app/templates/prometheusrules.yaml
diranged: Good catch... fixing.

charts/simple-app/templates/prometheusrules.yaml
diranged: @sabw8217 Thinking about this more ... I think this should be a global alarm. I can't think of a situation where a developer would have an autoscaling HPA in place, and want _different_ alarms than this?

charts/simple-app/Chart.yaml
```diff
@@ -2,13 +2,15 @@ apiVersion: v2
...
 dependencies:
+
   - name: istio-alerts
-    version: 0.1.1
-    repository: "file://../istio-alerts/"
-    condition: istio-alerts.enabled
+    alias: istioRules
+    version: 0.1.2
+    repository: https://k8s-charts.nextdoor.com
+    condition: istioRules.enabled
```
diranged, line 16: All of the `istio` changes above are in #98... so I will rebase this change once we merge #98, which will then let the tests pass.

charts/simple-app/Chart.yaml
```diff
@@ -2,13 +2,15 @@ apiVersion: v2
...
     email: matt@nextdoor.com
 dependencies:
+
   - name: istio-alerts
-    version: 0.1.1
-    repository: "file://../istio-alerts/"
-    condition: istio-alerts.enabled
+    alias: istioRules
+    version: 0.1.2
+    repository: https://k8s-charts.nextdoor.com
```
diranged, line 15: By using the user-facing repository endpoint, we no longer are forced to make a change  to all of the nested charts that depend on `istio-alerts` each time we update `istio-alerts`. Instead, each chart can be upgraded independently when we're ready. That seems better to me and less confusing.

charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -1,42 +1,50 @@
...
     nextdoor.com/source: {{ .Values.chart_source }}
 spec:
   groups:
-  - name: {{ .Release.Name }}.{{ .Release.Namespace }}.serviceRules
+  - name: {{ .Release.Namespace }}.{{ .Release.Name }}.{{ .Chart.Name }}.IstioServiceRules
     rules:
     {{ with .Values.serviceRules.http5XXMonitor }}
     {{ if .enabled }}
-    - alert: {{ $release.Name }}-5xx-Rate-Too-High
+    - alert: 5xx-Rate-Too-High
```
sabw8217, line 18: I like having the name of the service in the name of the alarm I get. `dev-changelog-5xx-rate-too-high` is a lot different than `us1-nextdoor-fe-fast-5xx-rate-too-high`!

charts/simple-app/templates/prometheusrules.yaml
diranged: I think it works pretty well.. This is the full name of the deployment. So of course someone could find a way to create another deployment that matches the prefix, but at some point we can't really prevent that. Here's the scenario I could imagine:

  * App1: "myapp"
  * App2: "myapp-async"

In the above case, you could find that the alerts created for "myapp" apply to both "myapp-.*" and "myapp-async-.*" pods. I don't see a good way to get around that. However, as I write this, I can think of another scenario that is more likely to be hit:

Developer intentionally (or not) creates a nearly 63-char long deployment name by setting `.fullnameOverride`: "really-long-string-that-is-too-long-and-may-be-truncated-for-sure". When this happens, the replicaset automatically truncates part of the name:

```
$  k get deployments,rs,pods
NAME                                                                              READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/really-long-string                                                1/1     1            1           6m7s
deployment.apps/really-long-string-that-is-too-long-and-may-be-truncated-for-su   1/1     1            1           5m32s

NAME                                                                                         DESIRED   CURRENT   READY   AGE
replicaset.apps/really-long-string-7f8689d4f                                                 1         1         1       6m7s
replicaset.apps/really-long-string-that-is-too-long-and-may-be-truncated-for-su-5cf866cd98   1         1         1       5m32s

NAME                                                                  READY   STATUS    RESTARTS   AGE
pod/really-long-string-7f8689d4f-tlk4x                                1/1     Running   0          6m7s
pod/really-long-string-that-is-too-long-and-may-be-truncated-fn4z5j   1/1     Running   0          2m1s
```

In this case, we need to handle the fact that the pod-name is going to be truncated to attach a 5-character string at the end... so we need to truncate the prefix for the `pod=~...` selector to 58 characters.

charts/istio-alerts/templates/service-prometheusrule.yaml
diranged: Sorry - we have two threads here... see https://github.com/Nextdoor/k8s-charts/pull/98#discussion_r783322249 for my response.

PR #98, opened 2022-01-12
[istio-alerts] Introduce .Chart.Name into Rule Names

diranged:
The motivation of introducing another field into .Chart.Name here is to
make it easier to identify where the "ServiceRules" are coming from -
without this, they appear to be natively part of the parent-chart
(simple-app for example). With the .Chart.Name in there, you can more
easily see that these come from a nested chart.

image

### DIFF
--- a/charts/istio-alerts/templates/service-prometheusrule.yaml
+++ b/charts/istio-alerts/templates/service-prometheusrule.yaml
@@ -1,7 +1,6 @@
-{{ $values     := .Values }}
-{{ $release    := .Release }}
-
-{{ if .Values.serviceRules.enabled }}
+{{- $values     := .Values }}
+{{- $release    := .Release }}
+{{- if .Values.serviceRules.enabled }}
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
@@ -13,14 +12,17 @@ spec:
   groups:
   - name: {{ .Release.Namespace }}.{{ .Release.Name }}.{{ .Chart.Name }}.IstioServiceRules
     rules:
-    {{ with .Values.serviceRules.http5XXMonitor }}
-    {{ if .enabled }}
+    {{- with .Values.serviceRules.http5XXMonitor }}
+    {{- if .enabled }}
     - alert: 5xx-Rate-Too-High
       annotations:
-        summary: {{`High rate of 5xx errors from {{$labels.destination_service_name}} in namespace {{$labels.namespace}}`}}
+        summary: >-
+          {{`{{$labels.destination_service_name}} is throwing too many 5xx errors (namespace: {{$labels.namespace}})`}}
         runbook_url: {{ $values.defaults.runbookUrl }}#5xx-Rate-Too-High
         description: >-
-          High rate of 5xx responses from the {{`$labels.destination_service_name`}} VirtualService in namespace {{`{{$labels.namespace}}`}}.
+          High rate of 5xx responses from the {{`{{$labels.destination_service_name}}`}} VirtualService
+          in namespace {{`{{$labels.namespace}}`}}.
+
       expr: >-
         sum by (destination_service_name) (
             increase(istio_requests_total{response_code=~"5.*", destination_service_namespace="{{ $release.Namespace }}"}[60s])
@@ -42,16 +44,16 @@ spec:
     {{- end }}
 
      # Adding latency alarm to trigger when request latency is above > threshold
-    {{ with .Values.serviceRules.highRequestLatency }}
-    {{ if .enabled }}
+    {{- with .Values.serviceRules.highRequestLatency }}
+    {{- if .enabled }}
     - alert: HighRequestLatency
       annotations:
-        summary: Processes experience latency above {{ .threshold }}.
+        summary: >-
+          {{`{{$labels.destination_service_name}}`}} avg request latencies are above {{ .threshold }}s!
         runbook_url: {{ $values.defaults.runbookUrl}} #HighRequestLatency
         description: >-
-          {{`{{ $value | humanizePercentage }} latency is above
-          namespace {{ $labels.namespace }} for container {{ $labels.container
-          }} in pod {{ $labels.pod }}.`}}
+          Average request latency of {{`{{ $value | humanizePercentage }}`}} is above threshold ({{ .threshold }}s)
+          in namespace {{`{{ $labels.namespace }}`}} for pod {{`{{ $labels.pod }}`}} (container: {{`{{ $labels.container }}`}}).
       expr: |-
         sum by (destination_service_name) (
           irate(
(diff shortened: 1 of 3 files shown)


charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -1,42 +1,50 @@
...
+        {{- with $values.defaults.additionalRuleLabels}}
+        {{ toYaml . | nindent 8}}
         {{- end }}
     {{- end }}
     {{- end }}
 
      # Adding latency alarm to trigger when request latency is above > threshold
     {{ with .Values.serviceRules.highRequestLatency }}
     {{ if .enabled }}
-    - alert: {{ $release.Name }}-HighRequestLatency
```
diranged, line 39: I removed this because we don't need the prefix here to identify where the alert came from in my opinion. If we want to help explain the application that is having the alert issue to the oncall-engineer, then we should do that through the `summary` field IMO.

sabw8217: I don't understand this at all. It's really helpful to _what service is being affected_ be front and center in an alert.

sabw8217: Compare with the alarm name for the 5xx alarm on the ALB:

`AlarmName: {'Fn::Sub': '${AWS::StackName}-5xxCount-Too-High'}`

diranged: So I totally agree that the developer getting paged wants as much context as possible. I just think the Alarm Name isn't necessarily the only way to do that. For example, take this PromethusRule:
```
    - alert: KubeQuotaAlmostFull
      annotations:
        description: 'Namespace {{ $labels.namespace }} is using {{ $value }} of its
          {{ $labels.resource }} quota. '
        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#KubeQuotaAlmostFull
        summary: Namespace quota is going to be full.
      expr: |-
        (
          kube_resourcequota{job="kube-state-metrics", type="used"}
            / ignoring(instance, job, type)
          (kube_resourcequota{job="kube-state-metrics", type="hard"} > 0)
        ) * 100 > 90 < 100
      for: 10m
      labels:
        severity: warning
```
here's the alert that a developer sees in PagerDuty:

Screen Shot 2022-01-12 at 10 01 14 AM

So the page comes through with the Alert Name and the `summary` field. The summary field already includes the service name and namespace... but we could reformat it so it's more clear:

```
summary: {{`{{$labels.destination_service_name}} throwing a high rate of 5xx errors`}}
```

charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -1,57 +1,71 @@
...
-    {{ with .Values.serviceRules.http5XXMonitor }}
-    {{ if .enabled }}
-    - alert: {{ $release.Name }}-5xx-Rate-Too-High
+    {{- with .Values.serviceRules.http5XXMonitor }}
+    {{- if .enabled }}
+    - alert: 5xx-Rate-Too-High
       annotations:
-        summary: {{`High rate of 5xx errors from {{$labels.destination_service_name}} in namespace {{$labels.namespace}}`}}
+        summary: >-
+          {{`{{$labels.destination_service_name}} is throwing too many 5xx errors (namespace: {{$labels.namespace}})`}}
```
diranged, line 20: Per our discussion, the `summary` field is where we want to have the information about the particular application and/or namespace. I've reformatted the summary to be every so slightly shorter but to make the service name front and center as the first word.

charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -1,57 +1,71 @@
...
         runbook_url: {{ $values.defaults.runbookUrl }}#5xx-Rate-Too-High
         description: >-
-          High rate of 5xx responses from the {{`$labels.destination_service_name`}} VirtualService in namespace {{`{{$labels.namespace}}`}}.
+          High rate of 5xx responses from the {{`{{$labels.destination_service_name}}`}} VirtualService
```
diranged, line 23: I reformatted this line and discovered bugs in the templating.

```
{{`$labels.destination_service_name`}}
```

needed to be
```
{{`{{$labels.destination_service_name}}`}}
```

The
```
{{`
```

tells Helm to stop processing this portion for templating... and we use that so that the final templated output is _a new template_ that will be processed by the alertmanager:

```
    - alert: 5xx-Rate-Too-High
      annotations:
        summary: >-
          {{$labels.destination_service_name}} is throwing too many 5xx errors (namespace: {{$labels.namespace}})
        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/istio-alerts/runbook.md#5xx-Rate-Too-High
        description: >-
          High rate of 5xx responses from the {{$labels.destination_service_name}} VirtualService
          in namespace {{$labels.namespace}}.
```
```

File: charts/istio-alerts/templates/service-prometheusrule.yaml [hunk 4]
```diff
@@ -1,57 +1,71 @@
...
       annotations:
-        summary: Processes experience latency above {{ .threshold }}.
+        summary: >-
+          {{`{{$labels.destination_service_name}}`}} avg request latencies are above {{ .threshold }}s!
         runbook_url: {{ $values.defaults.runbookUrl}} #HighRequestLatency
         description: >-
-          {{`{{ $value | humanizePercentage }} latency is above
-          namespace {{ $labels.namespace }} for container {{ $labels.container
-          }} in pod {{ $labels.pod }}.`}}
+          Average request latency of {{`{{ $value | humanizePercentage }}`}} is above threshold ({{ .threshold }}s)
```
diranged, line 55: Down here we ran into a different issue... we had used the template-escape code to escape the entire string. However, we had `{{ .threshold }}` in here that was not being processed... and that would have failed in processing by the AlertManager. With this code change, here's the final prometheusrule:

```yaml
     # Adding latency alarm to trigger when request latency is above > threshold
    - alert: HighRequestLatency
      annotations:
        summary: >-
          {{$labels.destination_service_name}} avg request latencies are above 0.5s!
        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/istio-alerts/runbook.md #HighRequestLatency
        description: >-
          Average request latency of {{ $value | humanizePercentage }} is above threshold (0.5s)
          in namespace {{ $labels.namespace }} for pod {{ $labels.pod }} (container: {{ $labels.container }}).
      expr: |-
        sum by (destination_service_name) (
          irate(
            istio_request_duration_milliseconds_bucket{reporter=~"destination",destination_service_namespace="default"}[1m]
          ) / 1000
        )
          > 0.5
      for: 15m
      labels:
        severity: warning
```

charts/istio-alerts/templates/service-prometheusrule.yaml
sabw8217: yo dawg I heard you like templates...

diranged: 😮‍💨

PR #102, opened 2022-01-27
[simple-app] Re-order HPA metrics

ericayin:
https://github.com/kubernetes/kubernetes/issues/74099

### DIFF
--- a/charts/simple-app/Chart.yaml
+++ b/charts/simple-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-app
 description: Default Microservice Helm Chart
 type: application
-version: 0.17.0
+version: 0.17.1
 appVersion: latest
 maintainers:
   - name: diranged
--- a/charts/simple-app/README.md
+++ b/charts/simple-app/README.md
@@ -2,7 +2,7 @@
 
 Default Microservice Helm Chart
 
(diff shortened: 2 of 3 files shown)


charts/simple-app/values.yaml
```diff
@@ -342,7 +342,7 @@ autoscaling:
   # -- Configures the HPA to target a particular CPU utilization percentage
   targetCPUUtilizationPercentage: 80
   # -- If supplied, configures the HPA to target a particular Memory utilization percentage
-  # targetMemoryUtilizationPercentage: 80
```
diranged, line 345: Eh. I think it should be uncommented, but also left blank. You don't generally want to autoscale on CPU and Memory in _most_ cases without lots of tuning. That's why it was left off here. It doesn't change the capability of the downstream chart from setting the value though.

ericayin: ooh yea that makes sense, i'll leave it as is

diranged (approved):
:+1: thank you!

charts/simple-app/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-app
 description: Default Microservice Helm Chart
 type: application
-version: 0.16.2
+version: 0.16.3
```
diranged, line 5: @ericayin oh rebase.. we're on 0.17+ now..

PR #103, opened 2022-01-28
[simple-app] Fix the ServiceMonitor and improve the Ports configuration

diranged:
I am working on using this chart to implement [YACE](https://github.com/nerdswords/yet-another-cloudwatch-exporter) - which is a service collects metrics from CloudWatch and exports so a ServiceMonitor can scrape them into Prometheus. I discovered a few bugs though with the way our `Service`, `ServiceMonitor` and `Deployment` resources were handling port configuration.

### DIFF
--- a/charts/simple-app/Chart.yaml
+++ b/charts/simple-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-app
 description: Default Microservice Helm Chart
 type: application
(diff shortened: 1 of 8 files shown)


gonzalo-nd (approved):
Nice! 
By the way, kudos on including Upgrade Notes.
As we do more changes we might want to formally introduce a Changelog file

PR #104, opened 2022-02-04
simple-app: Prefer latest curl image by default

smitelli:
The intent here is to fix this [build error](https://github.com/Nextdoor/hello-world-k8s.template/pull/11/checks?check_run_id=5069569863) by defaulting to a curl version that should support `--retry-connrefused`.


smitelli:
I saw the chart version lint error. I'll bump that if there are no other proposed changes.

PR #105, opened 2022-02-05
[stateful-app/simple-app/daemonset-app]: KMSSecret and Secrets management

dnguy078:
Abstracts handling kms-secrets into `simple-app`


charts/simple-app/templates/deployment.yaml
```diff
@@ -27,6 +27,9 @@ spec:
   template:
     metadata:
       annotations:
+        {{- if .Values.kms.secrets }}
+        checksum/config: {{ include (print $.Template.BasePath "/kmssecrets.yaml") . | sha256sum }}
+        {{- end}}
```
dnguy078, line 32: I don't believe K8 will restart the pods when secrets are updated. (https://kubernetes.io/docs/concepts/configuration/secret/#environment-variables-are-not-updated-after-a-secret-update) 

I added this to maybe trigger a restart of the pods if secrets are modified. (referenced https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) I'm exactly sure if this will work.

diranged: This is plenty good. Nit, add a space on line 32 before the `}}`

charts/simple-app/templates/deployment.yaml
```diff
@@ -241,8 +244,13 @@ spec:
                 command:
                   {{- toYaml . | nindent 18 }}
           {{- end }}
-          {{- with .Values.envFrom }}
+          {{- if or (.Values.envFrom) (.Values.kms.secrets) }}
```
diranged, line 247: No need for the parenthesis's I don't believe

charts/simple-app/values.yaml
```diff
@@ -177,6 +177,17 @@ envFrom: []
...
+# Configuration for managing KMS secrets. KMS Secrets Controller will decrypt each KMS secret and generate a K8 Secret Resource.
+# See https://github.com/h3poteto/kms-secrets for more details. For local development, sensitive environment variables can be injected through the `env` label.
+kms:
+  # -- (`Map`) Map of environment variables to KMS encrypted secrets.
+  secrets: {}
+  # Example:
+  # - SECRET_ENV: AQICAHhgSPI/REDACTED
+
+  # -- (String) AWS region where the KMS key is located
+  region: ""
```
diranged, line 189: Use `region:` here.. no need for `""`.

diranged (commented):
So this initial PR works fine - it adds in support for creating a `KMSSecret` resource in a consistent way, mapping it into the container, and triggering container restarts.

I am wondering if we want to go one step further to abstract away KMS here. Do we want something like:

```yaml
secrets:
  FOO_BAR: my plaintext secret
secretsEngine: plaintext
```

and alternatively
```yaml
secretsEngine: kms
secrets:
  FOO_BAR: AQIA...
```

In this example, either the `KMSSecret` or the `Secret` resource would be created and mounted into the container for the user... they wouldn't have to do the "if this, else this" thing. I am not sure though if we want to do this. @sabw8217 thoughts?

dnguy078:
> 

We could go one step further to abstract KMS. I was wondering what the use case for using the Native K8 secrets are. Aren't they just base64 encoded string values. We would not be able to securely store them in Git ? @diranged @sabw8217

dnguy078:
> So this initial PR works fine - it adds in support for creating a `KMSSecret` resource in a consistent way, mapping it into the container, and triggering container restarts.
> 
> I am wondering if we want to go one step further to abstract away KMS here. Do we want something like:
> 
> ```yaml
> secrets:
>   FOO_BAR: my plaintext secret
> secretsEngine: plaintext
> ```
> 
> and alternatively
> 
> ```yaml
> secretsEngine: kms
> secrets:
>   FOO_BAR: AQIA...
> ```
> 
> In this example, either the `KMSSecret` or the `Secret` resource would be created and mounted into the container for the user... they wouldn't have to do the "if this, else this" thing. I am not sure though if we want to do this. @sabw8217 thoughts?



> 

We could go one step further to abstract KMS. I was wondering what the use case for using the Native K8 secrets are. Aren't they just base64 encoded string values. We would not be able to securely store them in Git ? @diranged @sabw8217

charts/simple-app/templates/deployment.yaml
dnguy078: I simplified this a little further with `checksum/secrets: {{ toYaml .Values.kms.secrets | sha256sum }}`

diranged:
> 



> We could go one step further to abstract KMS. I was wondering what the use case for using the Native K8 secrets are. Aren't they just base64 encoded string values. We would not be able to securely store them in Git ? @diranged @sabw8217

@dnguy078 We'd never ever check in a real secret like that .. the motivation to support creating a `Secret` resource is purely for local dev and/or CI tests, where you are using bogus data anyways, but want to replicate the basic behavior of the chart. Let's chat tomorrow morning and I'll try to clarify.

charts/simple-app/values.yaml
```diff
@@ -177,6 +177,15 @@ envFrom: []
...
   #     name: my-config-vars
 
+# -- (`Map`) Map of environment variables to plaintext secrets or KMS encrypted secrets.
+secrets: {}
+
+# -- (String) AWS region where the KMS key is located
+kmsSecretsRegion:
+
+# -- (String) Secrets Engine determines the type of Secret Resource that will be created (`KMSSecret`, `Secret`). kms || plaintext are possible values.
+secretsEngine: plaintext
```
dnguy078, line 187: Open to switching this to being a bool flag. Maybe `kmsEncypted` flag ?

charts/simple-app/templates/deployment.yaml
```diff
@@ -253,6 +259,7 @@ spec:
                 name: {{ tpl (required ".name key must be set on secretRef" .name) $ }}
             {{- end }}
             {{- end }}
+            {{- end }}
```
diranged, line 262: I feel like this extra `end` is wrong... you've added a new `if...end` statement, but its all self contained... am I wrong?

charts/simple-app/templates/kmssecrets.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if and (.Values.secrets) (eq .Values.secretsEngine "kms") }}
```
diranged, line 1: Rename files to be singular - `kmssecret.yaml` and `secret.yaml`...

charts/simple-app/values.yaml
diranged: I think right now this is good.. there is another secret engine we use too sometimes, so we may choose to add that in here at some point.

diranged (commented):
just one question about a potential bug..

charts/simple-app/README.md
```diff
@@ -2,7 +2,7 @@
 
 Default Microservice Helm Chart
 
-![Version: 0.17.2](https://img.shields.io/badge/Version-0.17.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: latest](https://img.shields.io/badge/AppVersion-latest-informational?style=flat-square)
+![Version: 0.18.0](https://img.shields.io/badge/Version-0.18.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: latest](https://img.shields.io/badge/AppVersion-latest-informational?style=flat-square)
```
diranged, line 5: Edit the `README.md.gotmpl` and add in two sections..

1. An "upgade notes" sectin like https://github.com/Nextdoor/k8s-charts/blob/main/charts/simple-app/README.md.gotmpl#L15 that discusses what we're adding at a high level.

2. A section that discusses exactly how to use the settings...

diranged (changes requested):
Would you port this into daemonset-app and stateful-app as well please?

charts/simple-app/templates/deployment.yaml
dnguy078: This closes the `{{- with .Values.envFrom}}` on :251.

charts/simple-app/README.md
```diff
@@ -13,6 +13,13 @@ defaults for you like the Kubernetes [Horizontal Pod Autoscaler][hpa].
 
 ## Upgrade Notes
 
+### 0.17.x -> 0.18.x
+
+**New Feature: Secrets Management**
+
+You can now manage `Secret` and `KMS Secret` Resources through `Values.secrets`.
```
diranged, line 20: `KMSSecret` not `KMS Secret`

diranged (approved):
Great PR, thanks! :+1: to squash-and-merge...

PR #106, opened 2022-02-14
[Localstack] Add localstack custom helm chart

saansh45:


diranged:
@saansh45 This won't do much - it as in, it will not make this package available to our other charts via our repo... I think for now we should not bother putting this here and we'll add it next week when we create a new chart.

saansh45:
We don't need this anymore. Our changes have been updated in upstream chart https://github.com/localstack/helm-charts/pull/48

PR #116, opened 2022-02-24
[simple-app]: Add support for http rewrite in istio virtualservice configuration

saansh45:


charts/simple-app/values.yaml
```diff
@@ -291,6 +291,12 @@ virtualService:
   # for more details.
   corsPolicy: {}
 
+  # -- (`map`) If set, this will populate the rewrite setting for the
+  # VirtualService. See
+  # https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRewrite
+  # for more details.
+  rewrite: {}
```
diranged, line 298: Based on my read of https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRewrite - this will rewrite requests that match `uri.prefix` with the value of `rewrite.uri`... but our `uri.prefix` defaults to `/*`.

I sort of feel like the customization you are going for is going to have to be handled by creating a custom `VirtualService` resource in the applications own `templates/` directory. Nothing prohibits us from having multiple `VirtualService` entries too. So we could have one during the transition that is the "transition target", and a totally separate second one that is the "final target".

saansh45:
Closing this as we don't need to make these changes.

PR #117, opened 2022-03-09
[simple-app/stateful-app] Increase the Minimum Replica Count and set …

diranged:
…a PDB

Our devlopers are (rightfully) not setting `replicaCount` or
`podDisruptionBudget` configurations in staging because they haven't
felt the need to. However this means that during node replacements and
even during code rollouts we get the ArgoCD Notifications App reporting
that the service is "failed" just because something like the HPA wasn't
able to find metrics for a pod during the rollout.

It is reasonable to make our minimum default deployment include at least
2 pods, and have a `PodDisruptionBudget` created that allows only one
pod to be unavailable at a time. This more accurately reflects the
behaviors we have in production anyways.

### DIFF
--- a/charts/simple-app/Chart.yaml
+++ b/charts/simple-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-app
 description: Default Microservice Helm Chart
 type: application
-version: 0.19.0
+version: 0.20.0
 appVersion: latest
 maintainers:
   - name: diranged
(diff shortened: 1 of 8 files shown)


smitelli:
Looks reasonable overall. I don't know offhand how common it is for developers to omit setting these, but (assuming everybody uses the defaults) this would double the number of pods in the staging clusters. Is that -- or will that be -- a significant expense?

diranged:
I don't think expense will be a real issue here... and I think most developers are going to default to forgetting these settings entirely.

PR #120, opened 2022-03-15
[simple-app]: Parameterize datadog logging

saansh45:


charts/simple-app/templates/deployment.yaml
```diff
@@ -132,6 +132,9 @@ spec:
 
         {{- end }}
         {{- end }}
+        {{- if and .Values.datadog.enabled .Values.datadog.enableLogging .Values.datadog.service }}
+        ad.datadoghq.com/writer.logs: '[{"source":{{ .Values.datadog.service | quote}},"service":{{ .Values.datadog.service | quote}}]'
```
diranged, line 136: The container name needs to be parameterized here - “writer” in this case is what needs to be replaced with a template.

charts/simple-app/values.yaml
```diff
@@ -487,6 +487,11 @@ datadog:
   metricsToScrape:
     - '"*"'
 
+  # -- (`bool`) If true, then it will enable application logging to datadog.
+  # To enable datadog logging, datadog.service value must be set as it's
+  # being used as source and service name in datadog logging configuration.
```
diranged, line 492: Since we use “scrapeMetrics”, how about “scrapeLogs” or “collectLogs”?

charts/simple-app/templates/deployment.yaml
diranged: There are various other parameters that can be used (and we should encourage them) by the developer to parse the logs as they flow through. We should put in some of those here as parameters so that we can make it easy to do things like strip PII, skip noisy log lines, etc…

saansh45: Added source and service parameters. If not defined it will use default values (.Release.Name). Also added support for log processing rules. https://docs.datadoghq.com/agent/logs/advanced_log_collection/?tab=configurationfile

charts/simple-app/values.yaml
saansh45: Changed to scrapeLogs

charts/simple-app/templates/deployment.yaml
```diff
@@ -132,6 +132,9 @@ spec:
 
         {{- end }}
         {{- end }}
+        {{- if and .Values.datadog.enabled .Values.datadog.scrapeLogs.enabled }}
+        ad.datadoghq.com/{{ include "simple-app.name" . }}.logs: {{ include "simple-app.datadogScrapeLoggingConfig" . | toJson }}
+        {{- end }}
```
saansh45, line 137:
```    
        prometheus.io/path: /metrics
        ad.datadoghq.com/simple-app.logs: "\n- source:simple-app\n  service:simple-app\n  log_processing_rules:[map[name:exclude_datadoghq_users pattern:\\\\w+@datadoghq.com type:exclude_at_match]]"
      labels:
```

diranged: This doesn't look right ... the `[map[name:...` stuff looks like it isn't real JSON and will cause an issue. Also, can we make this easier to read by using yaml pipes to make this print out as multiline json?

eg instead of:
```yaml
ad.datadoghq.com/simple-app.logs: "\n- source:simple-app\n  ...
```

do
```yaml
ad.datadoghq.com/simple-app.logs: |-
  { "source": "simple-app",
    ...
  }
  ...
```

PR #121, opened 2022-03-21
[nd-common refactor] Migrate most of our Annotations/Labels into Nd-Common, and clarify Monitoring requirements

diranged:


charts/stateful-app/fg
```diff
@@ -0,0 +1,24 @@
+Pulling dependencies in...
```
smitelli, line 1: Probably don't need this. 😉

smitelli (approved):

charts/stateful-app/fg
diranged: @smitelli Hah! good catch..

PR #124, opened 2022-03-23
[simple-app]: Add support for datalog logging configuration

saansh45:
This will add functionality to scrape application logs by datadog. It adds support for providing tags (service and source) along with advanced log processing rules configuration https://docs.datadoghq.com/agent/logs/advanced_log_collection/?tab=docker . 

Screen Shot 2022-03-23 at 10 48 00 AM


charts/simple-app/values.yaml
```diff
@@ -489,6 +489,20 @@ datadog:
   metricsToScrape:
     - '"*"'
 
+  # -- (`map`) If true, then it will enable application logging to datadog.
+  # To enable datadog logging, datadog.service value must be set as it's
+  # being used as source and service name in datadog logging configuration.
+  scrapeLogs:
```
diranged, line 499: I think I'd rather flatten this out.... `.Values.datadog.scrapeLogs`, `.Values.datadog.scrapeLogsSource`, `.Values.datadog.scrapeLogsProcessingRules. Otherwise, I'd prefer everything to be under `.Values.datadog.scrapeLogs` in a single map like `.Values.datadog.scrapeLogs.enabled`, `.Values.datadog.scrapeLogs.processingRules`, `.Values.datadog.scrapeLogs.source`, etc..

charts/simple-app/values.yaml
```diff
@@ -489,6 +489,20 @@ datadog:
   metricsToScrape:
     - '"*"'
 
+  # -- (`map`) If true, then it will enable application logging to datadog.
```
diranged, line 496: Remove this comment because this commnt implies that `.Values.scrapeLogs` should be specifically set to true, when it's really a map.. When documenting a map, document the individual settings within it, and then the `helm_docs` process generates a nic readme..

charts/nd-common/templates/_datadog.tpl
```diff
@@ -28,6 +28,24 @@ ad.datadoghq.com/{{ .Chart.Name }}.instances: |-
...
+we add default values to it.
+*/}}
+
+{{- if and .Values.datadog.enabled .Values.datadog.scrapeLogs.enabled }}
+ad.datadoghq.com/{{ .Release.Name }}.logs: |-
+  [
+    {
+      "source": {{- default .Release.Name .Values.datadog.scrapeLogs.source | toJson }},
+      "service": {{- default .Release.Name .Values.datadog.service | toJson }},
+      "log_processing_rules": {{- .Values.datadog.scrapeLogsProcessingRules | toJson }}
```
diranged, line 44: For the users sake, it's probably worth running this through `tpl` as well...

charts/nd-common/templates/_datadog.tpl
```diff
@@ -28,6 +28,24 @@ ad.datadoghq.com/{{ .Chart.Name }}.instances: |-
...
+.Values.scrapeLogsProcessingRules map and convert into list of objects converted
+into json supported by datadog config. If source and service tag values not provided
+we add default values to it.
+*/}}
+
+{{- if and .Values.datadog.enabled .Values.datadog.scrapeLogs.enabled }}
+ad.datadoghq.com/{{ .Release.Name }}.logs: |-
+  [
+    {
+      "source": {{- default .Release.Name .Values.datadog.scrapeLogs.source | toJson }},
```
diranged, line 42: Why are we running a single blank string through `toJson`?

charts/nd-common/templates/_datadog.tpl
```diff
@@ -28,6 +28,24 @@ ad.datadoghq.com/{{ .Chart.Name }}.instances: |-
...
 
+{{/*
+This is datalog logging configuration. We take the .Values.scrapeLogs and
+.Values.scrapeLogsProcessingRules map and convert into list of objects converted
+into json supported by datadog config. If source and service tag values not provided
+we add default values to it.
+*/}}
+
+{{- if and .Values.datadog.enabled .Values.datadog.scrapeLogs.enabled }}
+ad.datadoghq.com/{{ .Release.Name }}.logs: |-
```
diranged, line 39: `.Release.Name` is incorrect... it might sometimes work, but the right thing to use is the `.Chart.Name` that we use to actually name the container:
image

charts/simple-app/values.yaml
saansh45: done

charts/nd-common/templates/_datadog.tpl
saansh45: done

charts/nd-common/templates/_datadog.tpl
saansh45: done

charts/nd-common/templates/_datadog.tpl
```diff
@@ -28,6 +28,24 @@ ad.datadoghq.com/{{ .Chart.Name }}.instances: |-
   ]
 {{- end }}
 
+{{/*
```
diranged, line 31: Use `{{- /*` here to wipe out any previous newlines (line 30) in the final rendered template.

charts/nd-common/templates/_datadog.tpl
```diff
@@ -28,6 +28,24 @@ ad.datadoghq.com/{{ .Chart.Name }}.instances: |-
...
+This is datalog logging configuration. We take the .Values.scrapeLogs and
+.Values.scrapeLogsProcessingRules map and convert into list of objects converted
+into json supported by datadog config. If source and service tag values not provided
+we add default values to it.
+*/}}
+
```
diranged, line 37: Remove this newline, OR replace line 36 with `*/ -}}` to wipe out this newline..

charts/simple-app/Chart.yaml
```diff
@@ -13,5 +13,5 @@ dependencies:
     repository: https://k8s-charts.nextdoor.com
     condition: istio-alerts.enabled
   - name: nd-common
-    version: 0.0.4
+    version: 0.0.5
```
diranged, line 16: You'll need to simultaneously update `daemonset-app` and `stateful-app` because both of those charts are also using `repository: file://../nd-common`. When we use the `file://...` reference, then you MUST update the version on all of the places where its used... I  hate it, but it's sort of a good forcing function to make sure we keep all the charts in-line.

charts/nd-common/templates/_datadog.tpl
```diff
@@ -28,6 +28,24 @@ ad.datadoghq.com/{{ .Chart.Name }}.instances: |-
...
+{{- if and .Values.datadog.enabled .Values.datadog.scrapeLogs.enabled }}
```
diranged, line 38: To test this out for basic function (not datadog function, but to make sure that the text is at least rendered), let's add some fake datadog settings into https://github.com/Nextdoor/k8s-charts/blob/main/charts/simple-app/ci/datadog-with-env-values.yaml and https://github.com/Nextdoor/k8s-charts/blob/main/charts/simple-app/ci/datadog-without-env-values.yaml that would cause this to get rendered... (oh and https://github.com/Nextdoor/k8s-charts/blob/main/charts/simple-app/values.local.yaml).. 

just makes it easier to catch errors if we are adding these annotations to our tests..

charts/nd-common/templates/_datadog.tpl
```diff
@@ -28,6 +28,24 @@ ad.datadoghq.com/{{ .Chart.Name }}.instances: |-
...
+ad.datadoghq.com/{{ .Chart.Name }}.logs: |-
```
diranged, line 39: Do you want to create a `nd-common.containerName` function in this PR while you are here? It'll be pretty simple:

```
# _common.tpl
{{- define "nd-common.containerName" -}}
{{ .Chart.Name }}
{{- end }}
```

but that will make it easier down the road for us to add customization.. for example
```

# _common.tpl
{{- define "nd-common.containerName" -}}
{{ default .Chart.Name .Values.containerName }}
{{- end }}
```

charts/nd-common/templates/_datadog.tpl
saansh45: fixed it

charts/nd-common/templates/_datadog.tpl
saansh45: done

charts/nd-common/templates/_datadog.tpl
```diff
@@ -27,7 +27,22 @@ ad.datadoghq.com/{{ .Chart.Name }}.instances: |-
     }
   ]
 {{- end }}
-
+{{- /*
+This is datalog logging configuration. We take the .Values.scrapeLogs and
```
smitelli, line 31: `datalog` -> `datadog` 😃

smitelli (approved):

saansh45:
It's duplicate of https://github.com/Nextdoor/k8s-charts/pull/127 . Closing this one.

PR #129, opened 2022-03-25
[nd-common]: Use quote instead of toJson function for datadog config …

saansh45:
…values


charts/nd-common/templates/_datadog.tpl
```diff
@@ -35,8 +35,8 @@ we add default values to it.
 ad.datadoghq.com/{{ include "nd-common.containerName" . }}.logs: |-
   [
     {
-      "source": {{ default .Chart.Name .Values.datadog.scrapeLogs.source | toJson }},
```
diranged, line 38: :+1:

diranged (approved):

PR #130, opened 2022-03-28
fix(nd-common): portName -> portNumber for

awgr:
.spec.podMetricsEndpoints.port

### DIFF
--- a/charts/nd-common/templates/_monitors.tpl
+++ b/charts/nd-common/templates/_monitors.tpl
@@ -52,7 +52,7 @@ spec:
     matchLabels:
       {{- include "nd-common.selectorLabels" . | nindent 6 }}
   podMetricsEndpoints:
-    - port: {{ .Values.monitor.portName }}
+    - port: {{ .Values.monitor.portNumber }}
       path: {{ .Values.monitor.path }}
       scheme: {{ .Values.monitor.scheme }}
       {{- with .Values.monitor.interval }}


awgr:
You can see how this value is consumed by simple-app [here](https://github.com/Nextdoor/k8s-charts/blob/0de77e43740ed3e8a5899ceb1215562a042dd39c/charts/simple-app/values.yaml#L401)

`monitor:
  # -- (`bool`) If enabled, PodMonitor resources for Prometheus Operator
  # are created or if `Values.istio.enabled` is `True`, then the appropriate
  # Pod Annotations will be added for the istio-proxy sidecar container to
  # scrape the metrics.
  enabled: true

  # -- (`string`) Name of the port to scrape for metrics - this is the name of
  # the port that will be exposed in your `PodSpec` for scraping purposes.
  portName: metrics

  # -- (`int`) Number of the port to scrape for metrics - this port will be
  # exposed in your `PodSpec` to ensure it can be scraped.
  portNumber: 9090`

charts/nd-common/templates/_monitors.tpl
```diff
@@ -52,7 +52,7 @@ spec:
     matchLabels:
       {{- include "nd-common.selectorLabels" . | nindent 6 }}
   podMetricsEndpoints:
-    - port: {{ .Values.monitor.portName }}
+    - port: {{ .Values.monitor.portNumber }}
```
diranged, line 55: According to the spec, this is supposed to be the name of the port: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#podmetricsendpoint

What is prompting you to try to change this?

PR #132, opened 2022-04-19
[stateful/simple/nd-comon] Add in NetworkPolicy resource

diranged:
The goal of this PR is to simplify the process of creating a `NetworkPolicy` resource that grants internal cluster communication access to the services inside these applications.


saansh45 (approved):
👍

PR #135, opened 2022-04-27
Fix simple-app ignoring the sidecar image tag

tongyliu:
The value for `proxySidecar.image.tag` was being ignored in `simple-app.proxyImageFqdn` leading to the main image tag always being used


saansh45 (approved):
👍

PR #136, opened 2022-05-03
[rollout-app] New Argo Rollouts based Deployment

diranged:
**What did I want?**

As we begin using [Argo Rollouts](https://argoproj.github.io/argo-rollouts/) - we need a new default application template that can create the `Rollout` objects, as well as the more customized `VirtualService` and `Service` resources. I wanted to enable us to begin experimenting with these deployment models in a simple way.

**What did I do?**

A newly created `rollout-app` chart has been created along with some customization to the Github Workflow. The workflow now detects if the `rollout-app` chart was updated, and if it was, we then install the Argo Rollouts controller/CRD for the tests. When the tests run, we verify that we can create both a `BlueGreen` and `Canary` rollout resource type. We are explicitly not testing actually triggering a rollout - we don't need to test the fundamental behavior of Argo, just that our chart basically works.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,15 +24,13 @@ jobs:
 
       - name: Set up Helm
         uses: azure/setup-helm@v1
-        with:
-          version: v3.4.0
 
       - uses: actions/setup-python@v2
         with:
           python-version: 3.7
 
       - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.0.1
+        uses: helm/chart-testing-action@v2.2.1
 
       - name: Run chart-testing (list-changed)
         id: list-changed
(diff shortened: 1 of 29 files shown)


.github/workflows/ci.yaml
```diff
@@ -24,15 +24,13 @@ jobs:
 
       - name: Set up Helm
         uses: azure/setup-helm@v1
-        with:
-          version: v3.4.0
```
diranged, line 28: This means we will automatically pick up the latest stable release going forward. Needed this because there were some bug fixes in Helm around the `ne` function.

.github/workflows/ci.yaml
```diff
@@ -24,15 +24,13 @@ jobs:
...
 
       - uses: actions/setup-python@v2
         with:
           python-version: 3.7
 
       - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.0.1
+        uses: helm/chart-testing-action@v2.2.1
```
diranged, line 33: Updated to use the latest `ct` tool

saansh45 (approved):
👍

PR #139, opened 2022-05-27
Add experimental recording rule to capture sample limit

smitelli:
I'm not sure if/how this is going to work. The idea is to add a recording rule to the common chart that constantly records the scalar value matching the configured scrape limit, so we have that metric in Prom and can include it in alerting queries.

The weirdness comes in the labeling. Here, these metrics are being created in "project" context, where there could potentially be more than one app in the same namespace with separate PodMonitors targeting a different subset of pods.

At the observability level, we have one alerting rule that is operating in "global" context, and aggregating down to the pod+namespace level. I'm having a hard time working out a set of metric labels that will accurately map "this grouping of pods" on one side to "this instance of an application chart" on the other.

@diranged Any thoughts on the labeling?


charts/nd-common/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: nd-common
 description: A helper chart used by most of our other charts
 type: library
-version: 0.0.10
+version: 0.0.11
```
smitelli, line 5: I know I have to bump all the other charts in the repo; didn't want to do that before being sure that this will work.

charts/nd-common/templates/_monitors.tpl
```diff
@@ -90,3 +93,38 @@ spec:
...
+  labels:
+    {{- include "nd-common.labels" . | nindent 4 }}
+    {{- with .Values.monitor.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  groups:
+  - name: {{ include "nd-common.fullname" . }}.monitorRules
+    rules:
+      - record: namespace:enforced_sample_limit
```
diranged, line 123: I see why you chose `namespace:`... but this is not a namespace-specific limit, instead its actually a `PodMonitor` specific limit. Since the current _enforced_ value is recorded as `prometheus_operator_prometheus_enforced_sample_limit`, I suggest `pod_monitor_sample_limit`.

diranged (commented):
So overall this works. I have a comment about the _name_ of the metric, but otherwise this is good:

image

charts/nd-common/templates/_monitors.tpl
smitelli: Changed.

diranged:
all good except for one comment..

PR #140, opened 2022-06-15
Flink Job: Expose nodeSelector and tolerations for the jobManager and taskManager

eugenend:
Flink Job: Expose nodeSelector and tolerations for the jobManager and taskManager


.github/workflows/ci.yaml
```diff
@@ -59,5 +59,9 @@ jobs:
         if: steps.list-changed.outputs.changed == 'true'
         run: make crds
 
+      - name: Wait
```
eugenend, line 62: There's a race condition and without this wait `ct` can't see the flink operator CRD

diranged (changes requested):
make sure you update values.yaml with appropriate new paramaeters and docstrings

charts/flink-job/values.yaml
```diff
@@ -200,6 +200,22 @@ flinkProperties:
   metrics.reporters: prom
   metrics.reporter.prom.class: org.apache.flink.metrics.prometheus.PrometheusReporter
 
+# -- (`map`) [Node selector] for Job Manager and Task Manager
+nodeSelector:
```
diranged, line 204: Don't actually set values here. You are just putting in an empty key as a pointer for anyone reading the docs:

```yaml
# -- (`map`) [Node selector] for Job Manager and Task Manager
nodeSelector:

# -- (`map`) [Tolerations] for Job Manager and Task Manager
tolerations:
```

eugenend: Done

charts/flink-job/templates/flinkjobcluster.yaml
```diff
@@ -53,7 +53,19 @@ spec:
...
     {{- end }}
-  # volumeClaimTemplates, initContainers, nodeSelector, tolerations, sideCars not exposed
+    # volumeClaimTemplates, initContainers, sideCars not exposed
+    {{- if .Values.nodeSelector }}
+    nodeSelector:
+      {{- with .Values.nodeSelector }}
+      {{- toYaml . | nindent 6 }}
+      {{- end }}
+    {{- end }}
+    {{- if .Values.tolerations }}
```
diranged, line 63: @eugenend You don't have to change this.. but in the interest of teaching you some helm tricks.. this can be shortened.

Long:
```yaml
    {{- if .Values.tolerations }}
    tolerations:
      {{- with .Values.tolerations }}
      {{- toYaml . | nindent 6 }}
      {{- end }}
    {{- end }}
```

Short:
```yaml
    {{- with .Values.tolerations }}
    tolerations:
      {{- toYaml . | nindent 6 }}
    {{- end }}
````

diranged (approved):
This is good... quick note I showed you how to shorten the code, if you want to take that approach.

charts/flink-job/templates/flinkjobcluster.yaml
eugenend: This is a good trick, thank you!

smitelli (approved):

PR #144, opened 2022-07-11
Added support for affinity setting for flinkjobcluster

eugenend:


charts/flink-job/templates/flinkjobcluster.yaml
```diff
@@ -70,6 +70,10 @@ spec:
       {{- with .Values.podLabels }}
       {{- toYaml . | nindent 6 }}
       {{- end }}
+    {{- with .Values.affinity }}
```
diranged, line 73: @eugenend Can you confirm that this key does exist in the Flink CRD?

eugenend: Yes, I'll close this PR for now

PR #154, opened 2022-07-18
[simple-app/stateful-app/daemonset-app] Use prometheus-alerts chart, use matchLabels again

diranged:
**Prometheus-Alerts**

I realized that at one point we started going down the path of breaking
out the different `PrometheusRules` into different charts (eg, moving
"Deployment" rules into the `simple-app` chart). The problem is that it
made it hard to reuse these rules elsewhere ... and we started to lose
sync of them between the `prometheus-rules` chart and the various
application charts.

Instead, this PR goes back to automatically including the
`prometheus-alerts` chart just like the `istio-alerts` chart is
included. The appropriate rules are disabled in each chart so that we
only create the rules that are necessary.

**Match Expressions vs Match Labels for TopologySpreadConstraints**

After speaking with Spot.io, they actually recommend we use `matchLabels` for `topologySpreadConstraints` ... they support it fully, and apparently have some known issues handling `matchExpressions` here.


charts/nd-common/templates/_topology_spread_constraints.tpl
```diff
@@ -41,15 +41,15 @@ TopologySpreadConstraint. The default value if not supplied is `1`.
...
-      {{- include "nd-common.selectorLabelsExpression" $ | nindent 6 }}
+    matchLabels:
+      {{- include "nd-common.selectorLabels" $ | nindent 6 }}
 {{- end -}}
 {{- if .Values.enableTopologySpread -}}
 - maxSkew: {{ default 1 .Values.topologySkew }}
   topologyKey: {{ default "topology.kubernetes.io/zone" .Values.topologyKey }}
   whenUnsatisfiable: DoNotSchedule
   labelSelector:
-    matchExpressions:
```
diranged, line 52: I am pretty sure the kubernetes-sigs descheduler does not support `matchExpressions` here, so it stopped working when we flipped over to this.

diranged:
@smitelli @saansh45 @LaikaN57 Please post-review. Merging this for now to ship out a fix.

smitelli:
:+1:

charts/nd-common/templates/_topology_spread_constraints.tpl
smitelli: :tired_face:

LaikaN57 (commented):
LGTM

PR #159, opened 2022-07-18
[simple/daemon/stateful] Revert to using in-template prometheusrules

diranged:
**What did I want?**

Unfortunately the use of nested-nested helm charts (ie, `myapp -> simple-app -> prometheus-alerts`) is fine when you only do it once. .. but when you run multiple versions of those nested sub charts, Helm squashes them into a common namespace and it causes all kinds of conflicts. This was the original reason I pulled the `prometheusRules` into the `simple-app` chart to begin with.

**What did I do?**

I finished the job of migrating all the custom prometheusRules into each of the daemons/stateful/simple-app charts. There is some duplication of effort here, but not a lot.

### DIFF
--- a/charts/daemonset-app/Chart.yaml
+++ b/charts/daemonset-app/Chart.yaml
@@ -2,17 +2,12 @@ apiVersion: v2
 name: daemonset-app
 description: Default DaemonSet Helm Chart
 type: application
-version: 0.6.1
+version: 0.7.0
 appVersion: latest
 maintainers:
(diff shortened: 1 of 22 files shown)


diranged:
**Proof: Simple-App**

image
image
image
image

diranged:
**Proof: Daemonset-App**
image
image
image

diranged:
**Proof: Stateful-APp Rules**

image
image
image

PR #165, opened 2022-07-26
[various] Add AWS Tags to Services and Ingresses

LaikaN57:
The "real change" here is just adding annotations (which get applies as AWS tags) to the following resources:

Services: `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: kubernetes_namespace={{ . Release.Namespace }}`
Ingresses: `alb.ingress.kubernetes.io/tags: kubernetes_namespace={{ . Release.Namespace }}`

This PR just spams these annotations on every service/ingress regardless of if it is actually going to create an AWS resource or not. It should be safe to just let this annotation hang if there is no AWS resource created.

### DIFF
--- a/charts/flink-operator/templates/metrics-service.yaml
+++ b/charts/flink-operator/templates/metrics-service.yaml
@@ -5,7 +5,6 @@ metadata:
     prometheus.io/port: "8443"
     prometheus.io/scheme: https
     prometheus.io/scrape: "true"
-    service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: kubernetes_namespace={{ .Release.Namespace }}
   name: {{ .Release.Name }}-controller-manager-metrics-service
   labels:
     control-plane: controller-manager
--- a/charts/flink-operator/templates/metrics-service.yaml
+++ b/charts/flink-operator/templates/metrics-service.yaml
@@ -5,7 +5,7 @@ metadata:
     prometheus.io/port: "8443"
     prometheus.io/scheme: https
     prometheus.io/scrape: "true"
-    service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: kubernetes_namespace={{ .Releases.Namespace }}
+    service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: kubernetes_namespace={{ .Release.Namespace }}
   name: {{ .Release.Name }}-controller-manager-metrics-service
   labels:
     control-plane: controller-manager
--- a/charts/flink-operator/templates/metrics-service.yaml
+++ b/charts/flink-operator/templates/metrics-service.yaml
@@ -5,6 +5,7 @@ metadata:
     prometheus.io/port: "8443"
     prometheus.io/scheme: https
     prometheus.io/scrape: "true"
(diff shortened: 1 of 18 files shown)


charts/flink-operator/templates/metrics-service.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
     prometheus.io/port: "8443"
     prometheus.io/scheme: https
     prometheus.io/scrape: "true"
+    service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: kubernetes_namespace={{ .Releases.Namespace }}
```
diranged, line 8: This only makes sense if we actually use a `Service` as a load balancer in our environment.. but we don't for reference. I don't mind adding it, but it's not going to do anything for us.

charts/nd-common/templates/_service.tpl
```diff
@@ -18,6 +18,8 @@ metadata:
   name: {{ default (include "nd-common.fullname" $) .Values.service.name }}
   labels:
     {{- include "nd-common.labels" $ | nindent 4 }}
+  annotations:
+    service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: kubernetes_namespace={{ .Releases.Namespace }}
```
diranged, line 22: I'm concerned that this implies that our `Service` objects are used to create load balancers in AWS .. but we don't do that anywhere. What problem are we trying to solve here?

diranged (changes requested):

charts/flink-operator/templates/metrics-service.yaml
LaikaN57: I put in the PR description that we are just spamming the annotation everywhere for now. I can change this to be conditional if we want.

charts/nd-common/templates/_service.tpl
LaikaN57: So this helper allows any type of service to be constructed. So we should add this annotation just in case someone does use Service.type=LB.

LaikaN57:
@diranged This PR is the "brute-force" method of just annotating everything and just letting the annotations not apply in certain situations. I can change this to use a condition, but wanted to check with you before I go put in those lines of code here.

diranged:
> @diranged This PR is the "brute-force" method of just annotating everything and just letting the annotations not apply in certain situations. I can change this to use a condition, but wanted to check with you before I go put in those lines of code here.

If that's the case - let's put in comments that indicate how these are not necessarily going to apply in a ton of cases, but they are safe defaults.

diranged:
> > @diranged This PR is the "brute-force" method of just annotating everything and just letting the annotations not apply in certain situations. I can change this to use a condition, but wanted to check with you before I go put in those lines of code here.
> 
> If that's the case - let's put in comments that indicate how these are not necessarily going to apply in a ton of cases, but they are safe defaults.

My concern here is someone copying-and-pasting and then expecting these to do something because they don't necessarily understand how they work.

LaikaN57:
https://github.com/Nextdoor/k8s-charts/compare/nd-common-0.0.16...nd-common-0.0.18 in `prometheus-alerts`.

PR #168, opened 2022-08-02
[all charts] Fix some bad ContainerWaiting/CrashLoop Rules

diranged:
The rules we had in place to catch Pods in a CrashLoopBackoff state were
too complicated and easily missed pods depending on the situation. A
much cleaner metric exists with a very clear selector for
`CrashLoopBackOff`, so I am rewriting the rules to use that.

https://www.circonus.com/2021/01/guide-to-monitoring-kubernetes-part-2-which-metrics-and-health-conditions-you-should-be-monitoring/


ronakts (approved):
don't know about any syntax issue otherwise lgtm.

PR #170, opened 2022-08-11
Flink Job: Expose volumeClaimTemplates for JobManager

eugenend:


eugenend:
here's where it's defined in the CRD https://github.com/Nextdoor/k8s-charts/blob/main/charts/flink-operator-crd/crds/flink-operator-crd.yaml#L3024

saansh45 (approved):
👍

PR #172, opened 2022-08-15
[WIP] Add support for an init container in simple-app

mrterry:
Add Values for an init_container

### DIFF
--- a/charts/simple-app/values.yaml
+++ b/charts/simple-app/values.yaml
@@ -122,6 +122,46 @@ readinessProbe:
     path: /
     port: http
 
+# An optional init container. This container starts and exits before the
+# rest of the pod containers start.  It is useful if there is setup work that
+# needs to happen before the real containers run, but is operationally
+# decoupled from what that container does.
+initContainer:
+  # -- (Boolean) Enables injecting a pre-defined reverse proxy sidecar
+  # container into the Pod containers list.
+  enabled: false
+
+  # -- (String) The name of the proxy sidecar container
+  name: proxy
+
+  image:
+    # -- (String) The Docker image name and repository for the sidecar
+    repository: nginx
+
+    # -- (String) The Docker tag for the sidecar
+    tag: latest
+
+    # -- (String) Always, Never or IfNotPresent
+    pullPolicy: IfNotPresent
+
+  # -- Environment Variables for the primary container. These are all run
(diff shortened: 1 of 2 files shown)


mrterry:
IMO the interface for this change is much more important that the details of the resources created.  @diranged @smitelli can I get a directional review of this before I start creating templates?

I don't think the init container needs a lot of configuration.  It needs env vars, volumes, and resources for sure, but I'm not sure needs much more. I don't need network access, but users might.  As such all things network are unimplemented in the interest of YAGNI.

As implemented, all the env vars, volumes, resources are entirely decoupled from the "main" container. We should consider having the init_container and main_container share configuration for env_vars, creds, volumes. I can't quite tell if this is a genius idea that will simplify configuration or a terrible idea that will turn into a mountain of shared passwords and mess.

charts/simple-app/values.yaml
```diff
@@ -122,6 +122,46 @@ readinessProbe:
     path: /
     port: http
 
+# An optional init container. This container starts and exits before the
```
diranged, line 125: So... I definitely don't want to create a dedicated configuration here for the initContainer. I would consier a single large parameter (list of maps)) that _is_ the `initContainers` value. The thing is that once you have one initContainer, who is to say you won't have two, or three... there's just too much customization available at that point.

I'd think:

```
# values.yaml
initContainers: []
```

```
# deployment.yaml
apiVersion: v1
...
spec:
  template:
    {{- with .Values.initContainers }}
    {{- tpl (toYaml .) $ | nindent XX }}
    {{- end }}
...
```

Even this though you should prototype on your laptop in your own chart and just see how it feels

mrterry: interesting.  as implemented, this has one bespoke initContainer, whose configuration is limited by the bits exposed in `values.yaml`.  This sounds like you would prefer a blind import from `values.yaml` with unlimited configuration.  Blindly passing `initContainers` through to the pod spec is certainly more powerful, but it is also a lot more rope for the user to hang themselves with.

diranged: yup... I feel like the moment we're talking about something as complex and powerful as an initContainer, we should release the reigns.

mrterry: ok.  see #173 for that approach.

PR #173, opened 2022-08-17
Expose initContainers to simple-app

mrterry:
Expose initContainers to simple-app

### DIFF
--- a/charts/simple-app/values.yaml
+++ b/charts/simple-app/values.yaml
@@ -47,6 +47,10 @@ args: []
 # https://kubernetes.io/docs/concepts/storage/volumes/.
 volumes: []
 
+# -- A list of init containers that are run before the main containers start
+# https://kubernetes.io/docs/concepts/workloads/pods/init-containers/
+initContainers: []
+
 # -- A stringified list of 'volumes' similar to the `Values.volumes` parameter,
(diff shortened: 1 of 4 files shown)


charts/simple-app/values.yaml
```diff
@@ -47,6 +47,10 @@ args: []
 # https://kubernetes.io/docs/concepts/storage/volumes/.
 volumes: []
 
+# -- A list of init containers that are run before the main containers start
+# https://kubernetes.io/docs/concepts/workloads/pods/init-containers/
+initContainers: []
```
mrterry, line 52: the annoying thing with doing it this way is that if there are any differences in `initContainers` among your various values variants, you have to entirely duplicate `initContainers` in each values file.  the issues I ran into in particular are
1. setting `image` explicitly rather than composing it from `repository` and `tag` (super annoying and will make it hard to integrate with existing deploy scripts).
2. setting smaller resource requirements for `local`.  this same issue would manifest for staging.

mrterry:
Landing this PR would give me enough rope to prototype Adroit using simple-app.  There is a separate discussion about whether we want Adroit to live on simple-app, but this would give me enough leash to start discovering other technical issues.

bwagner-nextdoor (approved):

PR #174, opened 2022-08-18
[simple-app]: Add minimum replica count check for PDB configuration

saansh45:
We have seen few times that replica count for deployment and pdb is configured as 1. It causes issue while scaling and draining the node. In order to configure PDB, deployment replica count should be set to minimum of 2. This PR adds that check. If replica count is less than 2, then helm will throw error. 

Proof it works

```
saansh45 ~/Code/k8s-charts/charts/simple-app (pdb_replica_count_check) $ make template
Pulling dependencies in...
helm dependency update .
walk.go:74: found symbolic link in path: /Users/angadpatil/Code/k8s-charts/charts/simple-app/ci/local-values.yaml resolves to /Users/angadpatil/Code/k8s-charts/charts/simple-app/values.local.yaml
Getting updates for unmanaged Helm repositories...
...Successfully got an update from the "https://k8s-charts.nextdoor.com" chart repository
Saving 2 charts
Downloading istio-alerts from repo https://k8s-charts.nextdoor.com
Deleting outdated charts
helm template --debug --values values.yaml --values values.local.yaml --namespace simple-app simple-app .
install.go:178: [debug] Original chart version: ""
install.go:195: [debug] CHART PATH: /Users/angadpatil/Code/k8s-charts/charts/simple-app

walk.go:74: found symbolic link in path: /Users/angadpatil/Code/k8s-charts/charts/simple-app/ci/local-values.yaml resolves to /Users/angadpatil/Code/k8s-charts/charts/simple-app/values.local.yaml

Error: execution error at (simple-app/templates/poddisruptionbudget.yaml:2:4): Deployment replica count can not be less than 2 in order to configure PDB. Please configure Replica count greater than 2.
helm.go:84: [debug] execution error at (simple-app/templates/poddisruptionbudget.yaml:2:4): Deployment replica count can not be less than 2 in order to configure PDB. Please configure Replica count greater than 2.
make: *** [template] Error 1
```

### DIFF
--- a/charts/simple-app/templates/poddisruptionbudget.yaml
+++ b/charts/simple-app/templates/poddisruptionbudget.yaml
@@ -1,7 +1,7 @@
+{{- if .Values.podDisruptionBudget -}}
 {{- if lt (.Values.replicaCount | int) 2 }}
 {{- fail "Deployment replica count can not be less than 2 in order to configure PDB. Please configure Replica count greater than 2." }}
 {{- end }}
-{{- if .Values.podDisruptionBudget -}}
 apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
--- a/charts/simple-app/templates/poddisruptionbudget.yaml
+++ b/charts/simple-app/templates/poddisruptionbudget.yaml
@@ -1,3 +1,6 @@
+{{- if lt (.Values.replicaCount | int) 2 }}
+{{- fail "Deployment replica count can not be less than 2 in order to configure PDB. Please configure Replica count greather than 2." }}
+{{- end }}
 {{- if .Values.podDisruptionBudget -}}
 apiVersion: policy/v1
 kind: PodDisruptionBudget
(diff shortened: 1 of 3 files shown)


charts/simple-app/templates/poddisruptionbudget.yaml
```diff
@@ -1,3 +1,6 @@
+{{- if lt (.Values.replicaCount | int) 2 }}
+{{- fail "Deployment replica count can not be less than 2 in order to configure PDB. Please configure Replica count greater than 2." }}
+{{- end }}
```
LaikaN57, line 3: This needs to go on the **inside** of the PDB check.

`{{- if .Values.podDisruptionBudget -}}`

LaikaN57 (changes requested):

charts/simple-app/templates/poddisruptionbudget.yaml
saansh45: It works but it should be executed only if PDB configuration is set. Good catch.

diranged: Should be greater than or equal to 2... not just greater than 2 (in your failure string)

saansh45: updated it.

LaikaN57 (approved):

diranged (approved):

PR #184, opened 2022-09-07
[simple-app] Create a unique PDB for each Deployment

diranged:
When running in deployment-per-AZ mode, it makes sense that the PDB should apply to each individual Deployment rather than all of the AZs as a whole.


mrterry (approved):
lgtm, but i don't have the right background to fully vet this.

PR #187, opened 2022-09-16
[all charts] Replace 1m rate functions with 5m

diranged:
When most of our pod metrics are collected every minute now, we can't run a rate() function over a 1m timestamp beacuse there aren't 2 datapoints. Changing this to 5m.


smitelli:
:+1:

PR #188, opened 2022-09-16
[simple/stateful/rollout] Update to istio-alerts chart 0.1.4

diranged:
This includes an important fix to the rate() setting causing some of our alarms to potentially not fire off.


smitelli (commented):
:+1:

PR #190, opened 2022-09-21
[rollout/simple/stateful/daemon] Add optional sidecar and init containers

stlava:
Similar to #173 but also allows for extra sidecars and use of template variables.

### DIFF
--- a/charts/simple-app/templates/deployment.yaml
+++ b/charts/simple-app/templates/deployment.yaml
@@ -140,12 +140,12 @@ spec:
         {{- with $.Values.volumes }}{{ toYaml . | nindent 8 }}{{ end }}
         {{- with $.Values.volumesString }}{{ tpl . $ | nindent 8 }}{{ end }}
       {{- end }}
+      {{- with $.Values.initContainers}}
       initContainers:
-        {{- with $.Values.initContainers}}
         {{- range $initContainer := index .}}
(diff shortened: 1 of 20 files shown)


charts/simple-app/templates/deployment.yaml
```diff
@@ -140,7 +140,18 @@ spec:
         {{- with $.Values.volumes }}{{ toYaml . | nindent 8 }}{{ end }}
         {{- with $.Values.volumesString }}{{ tpl . $ | nindent 8 }}{{ end }}
       {{- end }}
+      initContainers:
+        {{- with $.Values.initContainers}}
```
diranged, line 144: Move the `{{- with ...` to line 142, so that if no `initContainers` are provided, then we don't even create the `initContainer` key. Otherwise you end up with templating like this:

```yaml
...
initContainers:
containers:
  - ...
```

That kind of templating causes tools like ArgoCD to throw a fit in certain situations.

diranged (commented):
One change please + pls because you are adding a whole new feature, please update the `README.md.gotmpl` with a changelog entry describing the new feature.

Lastly ... since you're here.. would you consider doing this to the other charts so they are all on par?

stlava:
Comments have been addressed.

bwagner-nextdoor:
For clarity on this pr. Is #173 now redundant?

PR #194, opened 2022-10-19
[simple-app] Allow PDB creation if replica count is null

ericayin:
We assume that if somebody is explicitly setting the replica count to `null` and enabling PDBs, they know what they're doing and probably have custom autoscaling or something.

### DIFF
--- a/charts/simple-app/templates/deployment.yaml
+++ b/charts/simple-app/templates/deployment.yaml
@@ -337,8 +337,11 @@ spec:
 ---
 
 {{- if $.Values.podDisruptionBudget -}}
-{{- if lt ($.Values.replicaCount | int) 2 }}
+{{- if and $.Values.replicaCount (lt ($.Values.replicaCount | int) 2) }}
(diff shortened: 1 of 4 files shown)


charts/simple-app/templates/deployment.yaml
```diff
@@ -337,9 +337,12 @@ spec:
 ---
 
 {{- if $.Values.podDisruptionBudget -}}
-{{- if lt ($.Values.replicaCount | int) 2 }}
+{{- if and $.Values.replicaCount (lt ($.Values.replicaCount | int) 2) }}
```
diranged, line 340: Looks good - did you happen to test this to make sure it behaves as we expect?

ericayin: yea i tested it locally with the flask chart

diranged: excellent... sorry i just merged a change right before yours... can you rebase, bump the version, and then let's get it in..

PR #197, opened 2023-02-08
Updated CODEOWNERS for flink-related charts

eugenend:
Updated CODEOWNERS for flink-related charts


eugenend:
@corydolphin looks like groups/users mentioned in the `CODEOWNERS` need to have write permission to the repository

PR #203, opened 2023-05-07
[nd-common] Add ability to specify clusterIP for a service

perlow:
Setting `clusterIP: None` is needed to make a headless service, which is needed in some scenarios.


diranged (changes requested):
I'm not sure we really need to add this  because it's a fairly unique and common situation. That said, if you do, you need to update all of the charts that use this chart because we use `file://....` references in this repo.

Also, just double checking, but does `if .Values.service.clusterIP` work when it's set to `None`?

diranged:
Closing as abandoned for now - but not opposed to bringing this feature back.

PR #207, opened 2023-06-01
simple-workflow chart

masterginger:
This is a reusable Helm chart which helps create an Argo Workflow in a standardized way.
It uses [Argo Workflow DAG](https://argoproj.github.io/argo-workflows/walk-through/dag/) to define workflow steps and dependencies.

Please see the comments in `values.yaml` for more details.


.github/workflows/ci.yaml
```diff
@@ -63,6 +66,11 @@ jobs:
         run: |
           make -C charts/rollout-app install_rollouts
 
+      - name: Install Argo Rollouts Controller
```
bwagner-nextdoor, line 69: Should this be rollouts or simple workflow

charts/simple-workflow/templates/kmssecret.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if and (.Values.secrets) (eq .Values.secretsEngine "kms") }}
```
bwagner-nextdoor, line 1: It seems we are using this same template across several apps (simple app, stateful app, simple workflow) Could we move it to nd-common

masterginger: Agreed. We can have another PR to extract this out from the existing charts and put it to `nd-common`.

saansh45 (approved):

PR #209, opened 2023-06-01
just fix rollup time

bwagner-nextdoor:
We bucket in units of 1m, therefore this time must be greater. I greped for s] and 1m so I think this was the only case less than or equal to a minute in the repo


diranged (approved):

diranged:
:+1: thanks!

PR #212, opened 2023-06-07
BREAKING CHANGE: remove default readiness and liveness probes

tsingla1998:
Setting default `readinessProbe`s and `livenessProbe` is potentially helpful for web services but more and more we're building services that do not follow this pattern. Furthermore, we should encourage application writers to think concretely about what to set these probes since we've run into issues in the past with them being the same where they really should have been different.

Major version bump as this is a breaking change.

### DIFF
--- a/charts/simple-app/Chart.yaml
+++ b/charts/simple-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-app
 description: Default Microservice Helm Chart
 type: application
-version: 0.27.6
+version: 1.0.0
 appVersion: latest
 maintainers:
   - name: diranged
--- a/charts/simple-app/README.md.gotmpl
+++ b/charts/simple-app/README.md.gotmpl
@@ -12,6 +12,15 @@ defaults for you like the Kubernetes [Horizontal Pod Autoscaler][hpa].
 
 ## Upgrade Notes
 
+### 0.27.x -> 1.0.x
+
+**BREAKING: You need to explicitly set the `livenessProbe` and `readinessProbe`**
+
+In order to make the chart more flexible, we have removed the default values
+for the `livenessProbe` and `readinessProbe` parameters. You must now
+explicitly set them in your `values.yaml` file. If you do not, the chart will
+fail to install.
+
 ### 0.26.x -> 0.27.x
 
 **NEW: Optional sidecar and init containers**
--- a/charts/simple-app/templates/deployment.yaml
+++ b/charts/simple-app/templates/deployment.yaml
@@ -203,11 +203,11 @@ spec:
           {{- end }}
           {{- with $.Values.livenessProbe }}
           livenessProbe:
-            {{- toYaml . | nindent 12 }}
+            {{- toYaml tpl (required "livenessProbe is required" .) $ | nindent 12 }}
           {{- end }}
           {{- with $.Values.readinessProbe }}
           readinessProbe:
-            {{- toYaml . | nindent 12 }}
+            {{- toYaml tpl (required "readinessProbe is required" .) $ | nindent 12 }}
           {{- end }}
           resources:
             {{- toYaml $.Values.proxySidecar.resources | nindent 12 }}
--- a/charts/simple-app/values.yaml
+++ b/charts/simple-app/values.yaml
@@ -108,12 +108,12 @@ startupProbe:
 
 # -- A PodSpec container "livenessProbe" configuration object. Note that this
 # livenessProbe will be applied to the proxySidecar container instead if that
-# is enabled.
+# is enabled. This is **required**.
 livenessProbe:
 
 # -- A PodSpec container "readinessProbe" configuration object. Note that this
 # readinessProbe will be applied to the proxySidecar container instead if that
-# is enabled.
+# is enabled. This is **required**.
 readinessProbe:
 
 # -- (`ContainerPort[]`) A list of Port objects that are exposed by the
--- a/charts/simple-app/values.yaml
+++ b/charts/simple-app/values.yaml
@@ -110,17 +110,11 @@ startupProbe:
 # livenessProbe will be applied to the proxySidecar container instead if that
 # is enabled.
 livenessProbe:
-  httpGet:
-    path: /
-    port: http
 
 # -- A PodSpec container "readinessProbe" configuration object. Note that this
 # readinessProbe will be applied to the proxySidecar container instead if that
 # is enabled.
 readinessProbe:
-  httpGet:
-    path: /
-    port: http
 
 # -- (`ContainerPort[]`) A list of Port objects that are exposed by the
 # service. These ports are applied to the main container, or the proxySidecar
--- a/charts/simple-app/README.md
+++ b/charts/simple-app/README.md
@@ -2,7 +2,7 @@
 
 Default Microservice Helm Chart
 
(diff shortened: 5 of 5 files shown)


charts/simple-app/values.yaml
```diff
@@ -110,17 +110,11 @@ startupProbe:
 # livenessProbe will be applied to the proxySidecar container instead if that
 # is enabled.
 livenessProbe:
-  httpGet:
```
diranged, line 113: I don't think this is the answer we want... there are far more stacks that _use this_ than that do _not_ use this. Without a default liveness/readiness probe, we're relying on engineers to figureo ut on their own that they need these.

Please do not merge this for now. This will require internal discussion.

diranged (changes requested):

charts/simple-app/values.yaml
stlava: > we're relying on engineers to figureo ut on their own that they need these.

@tsingla1998's point is that these two are critical and it should require developers to think and be very explicit about how liveness and readiness is treated for their application.

diranged: So - I could imagine a scenario where we make the chart _hard fail_ if you don't put in a `livenessProbe` and `readinessProbe`? But having them simply _not there_ by default doesn't feel right to me. What do you think about that?

tsingla1998: Agreed and done.

tsingla1998:
Side note... should we put `main` branch protections for this repo? If I see a big green button telling me I can merge, I'm inclined to click it. 😂

diranged:
@saansh45, @LaikaN57,
  What do you guys think of this change? I think I generally approve .. but it changes the developer experience with the chart noticeably (hence the major version change).

saansh45:
> @saansh45, @LaikaN57, What do you guys think of this change? I think I generally approve .. but it changes the developer experience with the chart noticeably (hence the major version change).



> @saansh45, @LaikaN57, What do you guys think of this change? I think I generally approve .. but it changes the developer experience with the chart noticeably (hence the major version change).

@diranged I agree that it will change developer experience and it's breaking change for them once they update to latest simple-app chart. But I think this is important change as it gives developers ability to define probes as per their needs. And with your recent push to make it as required (or should throw errors), this makes more sense and useful. Only thing I would say is that we need announce more broadly these breaking changes to developers.

saansh45:
@tsingla1998 can you please update https://github.com/Nextdoor/k8s-charts/blob/main/charts/simple-app/README.md.gotmpl ?

saansh45 (approved):

charts/simple-app/values.yaml
LaikaN57, line 113: Convert type of `livenessProbe` from `null` to empty object `{}`. I do not remember exactly why/how this works, but I do remember that we do not want to do multi-typing.

Let me also ask #cloudeng about this one too as I am not totally sure what I am remembering here is applicable to this specific scenario.

charts/simple-app/values.yaml
```diff
@@ -110,17 +110,11 @@ startupProbe:
...
-    path: /
-    port: http
 
 # -- A PodSpec container "readinessProbe" configuration object. Note that this
 # readinessProbe will be applied to the proxySidecar container instead if that
 # is enabled.
 readinessProbe:
-  httpGet:
```
LaikaN57, line 121: Same. Type conversion.

charts/simple-app/README.md.gotmpl
```diff
@@ -12,6 +12,15 @@ defaults for you like the Kubernetes [Horizontal Pod Autoscaler][hpa].
 
 ## Upgrade Notes
 
+### 0.27.x -> 1.0.x
+
+**BREAKING: You need to explicitly set the `livenessProbe` and `readinessProbe`**
```
LaikaN57, line 17: Add required param text to values file.

LaikaN57 (changes requested):

charts/simple-app/values.yaml
tsingla1998: I copied `startupProbe` above.

charts/simple-app/templates/deployment.yaml
```diff
@@ -203,11 +203,11 @@ spec:
           {{- end }}
           {{- with $.Values.livenessProbe }}
           livenessProbe:
-            {{- toYaml . | nindent 12 }}
+            {{- toYaml tpl (required "livenessProbe is required" .) $ | nindent 12 }}
```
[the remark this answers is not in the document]
tsingla1998: The k8s onboarding experience is currently not great due to this default setting in particular. I don't know why this would be any different/more confusing than all the other places that use `required`

charts/simple-app/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-app
 description: Default Microservice Helm Chart
 type: application
-version: 0.27.6
+version: 1.0.0
```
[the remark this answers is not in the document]
tsingla1998: I have no strong feelings on this. Please provide consistent guidance.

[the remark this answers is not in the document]
corydolphin: In my experience, semver and "marketing" versions are often in tension. I would prefer we provide clarity of the semantic versioning (e..g this is a breaking change).

charts/simple-app/README.md.gotmpl
corydolphin: This looks duplicative of the other comment on values.yaml.

charts/simple-app/values.yaml
LaikaN57: ⛔️ Blocking comment.

charts/simple-app/values.yaml
LaikaN57: ⛔️ Blocking comment.

charts/simple-app/README.md.gotmpl
LaikaN57: ⛔️ Blocking comment.

I feel that we should either add more text to the parameter in the values file explicitly stating that this is a required field and/or we should upgrade the required text output on failure to be a bit more human descriptive/helpful/breadcrumbs.

I feel strongly about this for OOBE reasons. Just because our current OOBE is not great does not mean that we should continue to allow not-so-great pattern matching. Lets strive to improve ourselves in _this_ PR. 🤷🏼 idk just my take. Didn't feel it was that hot as it should be quick docfix.

LaikaN57 (commented):
Add indicators for blocking comments.

charts/simple-app/README.md.gotmpl
tsingla1998: Done.

charts/simple-app/Chart.yaml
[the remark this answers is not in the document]
tsingla1998: Going to stick with SemVer.

charts/simple-app/templates/deployment.yaml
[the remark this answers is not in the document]
tsingla1998: Fixed by adding required in documentation for values.yaml file

charts/simple-app/values.yaml
tsingla1998: going to leave as is and all probes can be updated in a followup PR following best typing practices.

charts/simple-app/values.yaml
tsingla1998: going to leave as is and all probes can be updated in a followup PR following best typing practices.

PR #214, opened 2023-07-08
feat(charts/rollout-app): Support EKS 1.26 changes to `autoscaling` API group

smitelli:
The deprecated v2beta2 HorizontalPodAutoscaler API is removed in EKS 1.26 and replaced with v2.


pribeiro-nextdoor (approved):

LaikaN57 (approved):

LaikaN57:
We will need to update all usages of this chart.

smitelli:
Already handled in #222.

smitelli:
> We will need to update all usages of this chart.

The only usage I was able to find is konductor, which has been updated and deployed.

PR #215, opened 2023-07-10
feat(charts/simple-workflow): Support 'verifyOnly' and 'revision'

masterginger:
This PR introduced two new functionalities to `simple-workflow` chart:
* `verifyOnly` which allows `simple-workflow` to only perform verification on an Argo App which is deployed by other pipelines, without trying to create/update the app itself.
* `revision` which allows `simple-workflow` to verify the Argo App based on the revision the app is synced with, instead of the `targetRevision` field.

### DIFF
--- a/charts/simple-workflow/Chart.yaml
+++ b/charts/simple-workflow/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-workflow
 description: Default Argo Workflow Helm Chart
 type: application
-version: 0.0.2
+version: 0.0.3
 appVersion: latest
 maintainers:
   - name: masterginger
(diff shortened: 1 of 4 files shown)


jeremy-fisher-nd (approved):
lgtm! It looks like this is just porting over the changes [here](https://github.com/Nextdoor/adroit/blob/e553f91591255e544f93cd337e226b3486362d22/charts/simple-workflow/templates/workflow-pipeline.yaml#L31) right

masterginger:
> lgtm! It looks like this is just porting over the changes [here](https://github.com/Nextdoor/adroit/blob/e553f91591255e544f93cd337e226b3486362d22/charts/simple-workflow/templates/workflow-pipeline.yaml#L31) right

Right, then we will update `adroit` to use the `simple-workflow` from k8s-charts repo (this repo)

PR #216, opened 2023-07-11
feat(simple-app,rollout-app,stateful-app): support multiple namespaces for gateways

diranged:
The existing code made it difficult (impossible) to support moving from a `Gateway` in `ns1` to another one in `ns2`. It also assumed that `Gateway` objects lived in `istio-system` which is specifically against the [best practices in the Istio guide](https://istio.io/latest/docs/setup/additional-setup/gateway/#deploying-a-gateway) where they are recommended to live in a dedicated namespace.

This PR introduces a **breaking change** to the `rollout-app`, `statefull-app`, and `simple-app` charts to reformat the way the `VirtualService` object's `.spec.gateways` field is managed. I have introduced a loud `fail()` message to ensure that you cannot upgrade the chart without making the appropriate change.


.github/workflows/stale.yaml
```diff
@@ -0,0 +1,21 @@
+# https://github.com/marketplace/actions/stale-branches
```
diranged, line 1: Added this in ... because: 
image

PR #218, opened 2023-07-13
feat(charts/istio-endpoints): Scope down the service captures

diranged:


diranged:
Also fixing `port` which should be `portNumber`:
```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(EnvoyFilter.spec.configPatches[0].match.cluster): unknown field "port" in io.istio.networking.v1alpha3.EnvoyFilter.spec.configPatches.match.cluster
```
https://istio.io/latest/docs/reference/config/networking/envoy-filter/#EnvoyFilter-ClusterMatch

PR #227, opened 2023-08-16
Adding support for batch job/s in simple-workflow

kavinnd:
What is this ?
- This PR adds on to the  existing simple-workflow that checks on argocd app health to also look for one or many `kind: job` 's health within the same argocd app.

Why do we need this ?
- Currently, simple-workflow requires a `kind: Job` to not be an [argo hook](https://argo-cd.readthedocs.io/en/stable/user-guide/resource_hooks/).
- Having a job as hook is beneficial because we can decide if we want to run is Pre, Post or During sync.
- This change will enable argocd applications to have hooks that can run as part of their sync and ensure they are healthy too.

### DIFF
--- a/charts/simple-workflow/templates/workflow-pipeline.yaml
+++ b/charts/simple-workflow/templates/workflow-pipeline.yaml
@@ -195,7 +195,7 @@ spec:
               {{- end }}
        {{- range $jobName, $jobConfig := $config.jobs }}
         - name: verify-job-synced-and-healthy-{{ $name }}
-          depends: verify-app-synced-and-healthy-{{ $name }}.Succeeded
+          depends: verify-job-healthy-{{ $name }}.Succeeded
           template: verify-job-{{ $jobName }}
           arguments:
             parameters:
--- a/charts/simple-workflow/templates/workflow-pipeline.yaml
+++ b/charts/simple-workflow/templates/workflow-pipeline.yaml
@@ -195,7 +195,7 @@ spec:
               {{- end }}
        {{- range $jobName, $jobConfig := $config.jobs }}
         - name: verify-job-synced-and-healthy-{{ $name }}
-          depends: verify-job-healthy-{{ $name }}.Succeeded
+          depends: verify-app-synced-and-healthy-{{ $name }}.Succeeded
           template: verify-job-{{ $jobName }}
           arguments:
             parameters:
--- a/charts/simple-workflow/templates/workflow-pipeline.yaml
+++ b/charts/simple-workflow/templates/workflow-pipeline.yaml
@@ -129,6 +129,23 @@ spec:
           {{- with $config.metadata }}
           {{- tpl (toYaml .) $ | nindent 10 }}
           {{- end }}
+  {{- range $jobName, $jobConfig := $config.jobs }}
+  - name: verify-job-{{ $jobName }}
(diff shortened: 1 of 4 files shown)


charts/simple-workflow/templates/workflow-pipeline.yaml
```diff
@@ -176,6 +193,16 @@ spec:
               {{- else }}
               value: status.sync.status == Synced, status.health.status == Healthy, status.sync.comparedTo.source.targetRevision == {{ $.Values.app.targetRevision }}
               {{- end }}
+       {{- range $jobName, $jobConfig := $config.jobs }}
+        - name: verify-job-synced-and-healthy-{{ $name }}
+          depends: verify-app-synced-and-healthy-{{ $name }}.Succeeded
```
masterginger, line 198: Why does it depend on app-synced-and-healthy ?

charts/simple-workflow/templates/workflow-pipeline.yaml
```diff
@@ -176,6 +193,16 @@ spec:
...
+          template: verify-job-{{ $jobName }}
+          arguments:
+            parameters:
+              - name: successCondition
+                value: status.health.status == Healthy
```
masterginger, line 203: Do we need both synced and healthy?

charts/simple-workflow/templates/workflow-pipeline.yaml
kavinnd: duh, copy pasta err.

charts/simple-workflow/templates/workflow-pipeline.yaml
kavinnd: In case of a job there is no Sync actually.

Screenshot 2023-08-16 at 3 36 10 PM

charts/simple-workflow/templates/workflow-pipeline.yaml
```diff
@@ -176,6 +193,16 @@ spec:
               {{- else }}
               value: status.sync.status == Synced, status.health.status == Healthy, status.sync.comparedTo.source.targetRevision == {{ $.Values.app.targetRevision }}
               {{- end }}
+       {{- range $jobName, $jobConfig := $config.jobs }}
+        - name: verify-job-synced-and-healthy-{{ $name }}
+          depends: verify-job-healthy-{{ $name }}.Succeeded
```
masterginger, line 198: Where is this dependency defined?

kavinnd: actually, we ll wait for the app to be healthy before checking if the job is healthy. This should be true with any type of Hooks.  reverting my change that I thought was a copy past err.

charts/simple-workflow/templates/workflow-pipeline.yaml
masterginger: If there's a logical dependency between app and the job, job config should be a sub dictionary of app config.

kavinnd: it is actually. checkout https://github.com/Nextdoor/k8s-charts/pull/227/files#diff-15e2fdd4191db2b63c40fa4fb6dd7fab5efb400112e6b51f61fdfb66f2d6048c

masterginger (approved):
Great! Saw "jobs" under the app config.

PR #229, opened 2023-08-21
fix(charts/simple-workflow): Update the successCondition of a job

kavinnd:
- The batch Job has a different status than an application. 

Screenshot 2023-08-21 at 3 18 34 PM

- I updated the success condition to check for status.conditions[0].type == Complete

- Tested by running 
```
k wait --for=jsonpath='{.status.conditions[0].type}'=Complete  job.batch/schema-sync
job.batch/schema-sync condition met

```

### DIFF
--- a/charts/simple-workflow/Chart.yaml
+++ b/charts/simple-workflow/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-workflow
 description: Default Argo Workflow Helm Chart
 type: application
-version: 0.0.8
+version: 0.0.9
 appVersion: latest
 maintainers:
   - name: masterginger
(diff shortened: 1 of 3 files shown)


masterginger (commented):
Is "1" just representing "true" or is it a real number like "number of successful runs"?

kavinnd:
> Is "1" just representing "true" or is it a real number like "number of successful runs"?

Good question. It is the `number of successful runs`. I think a better check would be to either make this configurable ? 
or I think a better check would be on status.conditions.type == 'Complete'

kavinnd:
> > Is "1" just representing "true" or is it a real number like "number of successful runs"?
> 
> Good question. It is the `number of successful runs`. Ref - https://kubernetes.io/docs/concepts/workloads/controllers/job/#backoff-limit-per-index
> 
> I think a better check would be to either make this configurable ? or I think a better check would be on status.conditions[0].type == 'Complete'

Updated the condition Bian @masterginger.

masterginger (approved):

PR #230, opened 2023-08-29
Add version 1.6 of native flink-kubernetes-operator to k8s-charts

sbrother:
The most up-to-date version of the operator that exists on external hosting is 1.4, which has a bug that causes failure in our use case when using `watchNamespaces`. This commit contains `release-1.6` from the upstream repo (https://github.com/apache/flink-kubernetes-operator/tree/main/helm/flink-kubernetes-operator). It does *not* include the crds, which we are managing separately.


sbrother:
Closing as we are just using https://archive.apache.org/dist/flink/flink-kubernetes-operator-1.6.0/ with skipCrds set to true in Argo.

PR #243, opened 2023-11-21
fix(charts/daemonset-app): Remove default probes

scohen-nd:
Remove default probes to fix an issue of multiple probes.

### DIFF
--- a/charts/daemonset-app/values.yaml
+++ b/charts/daemonset-app/values.yaml
@@ -98,17 +98,17 @@ volumes: []
 # -- A PodSpec container "startupProbe" configuration object. Note that this
 # startupProbe will be applied to the proxySidecar container instead if that
 # is enabled.
-startupProbe: {}
+startupProbe: null
 
 # -- A PodSpec container "livenessProbe" configuration object. Note that this
 # livenessProbe will be applied to the proxySidecar container instead if that
 # is enabled.
-livenessProbe: {}
+livenessProbe: null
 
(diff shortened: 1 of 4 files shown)


charts/daemonset-app/values.yaml
```diff
@@ -98,23 +98,17 @@ volumes: []
 # -- A PodSpec container "startupProbe" configuration object. Note that this
 # startupProbe will be applied to the proxySidecar container instead if that
 # is enabled.
-startupProbe:
+startupProbe: {}
```
corydolphin, line 101: In `simple-app` we set these to null. should we do that for consistency?

scohen-nd: Done

diranged (commented):
Since this is a backwards incompatible change - can you bump the version to 1.14 and put in a README section in https://github.com/Nextdoor/k8s-charts/blob/main/charts/daemonset-app/README.md.gotmpl#L15-L28?

scohen-nd:
> can you bump the version to 1.14 and put in a README section i

Done

diranged:
@scohen-nd I don't see a readme update.. check the readme.md.gotmpl file for examples.

diranged (approved):

scohen-nd:
> @scohen-nd I don't see a readme update.. check the readme.md.gotmpl file for examples.

Addressed that as well, my bad.

PR #249, opened 2023-11-29
feat(charts/simple-app): support addition of custom alerts to default prometheusrules

schahal:
## Motivation

Right now if someone wants to add their own rules, they really have no choice but to add a `templates/` file for `kind: PrometheusRule`. This should provide a simple extension/interface to add new rules.

## Testing

(1) Before change, ran a `helm template test . -f values.yaml > /tmp/orig.yaml`

(2) After change, updated `values.yaml` to be:

```
prometheusRules:
  ...
  additionalCustomRules: []
```

.. and redirected template render output to `/tmp/orig2.yaml`

**As expected**, no diff when running `diff /tmp/orig.yaml /tmp/orig2.yaml`

(3) Added 2 dummy alert rules to `values.yaml`:

```
prometheusRules:
  ...
  additionalCustomRules:
  - alert: FoobarNotUp
    annotations:
      summary: Job foobar is not up
      runbook_url: https://example.com
      description:
    expr: up{job="foobar"} == 0
    for: 10m
    labels:
      severity: critical
  - alert: Foobar2NotUp
    annotations:
      summary: Job foobar2 is not up
      runbook_url: https://example.com
      description:
    expr: up{job="foobar2"} == 0
    for: 10m
    labels:
      severity: okay
```

... and redirected output to /tmp/new.yaml.

Got expected diff:

```diff
--- /tmp/orig.yaml	2023-11-29 14:20:03
+++ /tmp/new.yaml	2023-11-29 14:25:12
@@ -430,6 +430,26 @@
       for: 15m
       labels:
         severity: warning
+  - name: default.test.simple-app.CustomRules
+    rules:
+    - alert: FoobarNotUp
+      annotations:
+        description: null
+        runbook_url: https://example.com
+        summary: Job foobar is not up
+      expr: up{job="foobar"} == 0
+      for: 10m
+      labels:
+        severity: critical
+    - alert: Foobar2NotUp
+      annotations:
+        description: null
+        runbook_url: https://example.com
+        summary: Job foobar2 is not up
+      expr: up{job="foobar2"} == 0
+      for: 10m
+      labels:
+        severity: okay
 ---
 # Source: simple-app/templates/servicemonitor.yaml
 apiVersion: monitoring.coreos.com/v1
```

### DIFF
--- a/charts/daemonset-app/Chart.yaml
+++ b/charts/daemonset-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: daemonset-app
 description: Default DaemonSet Helm Chart
 type: application
-version: 0.14.0
+version: 0.14.1
 appVersion: latest
 maintainers:
   - name: diranged
--- a/charts/daemonset-app/README.md
+++ b/charts/daemonset-app/README.md
@@ -2,7 +2,7 @@
 
 Default DaemonSet Helm Chart
 
-![Version: 0.14.0](https://img.shields.io/badge/Version-0.14.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: latest](https://img.shields.io/badge/AppVersion-latest-informational?style=flat-square)
+![Version: 0.14.1](https://img.shields.io/badge/Version-0.14.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: latest](https://img.shields.io/badge/AppVersion-latest-informational?style=flat-square)
 
 [statefulsets]: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/
 [hpa]: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/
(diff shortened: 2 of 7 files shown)


diranged (commented):
@schahal,
  Out of curiosity - why would we do this, rather than just having developers create their own `PrometheusRule` template in their application helm chart?

schahal:
A few reasons that came to mind, mainly his will abstract away the "helm-isms" from the app developer:

* They wouldn't have to worry about writing own `kind: PrometheusRule` template
* The resource would be annotated/named consistently with simple-app's template schematics
* If wanting to disable/tweak specific custom alert for a specific environment, the template within their application chart wouldn't be static (leading to more churn to get a functioning template).

My _**gut**_ feeling was this could (would) be an easy/quick win. But if it seems out-of-band with the intention of this chart (which I may very well not have a good grasp on), and that this is currently not a real pain for app developers, I'm more than happy to defer raw `PrometheusRule` template files directly in app charts.

(also, the gosh-darn helm-docs ci keeps failing!)

schahal:
Closing this. After review, there are a "can-of-worms" that can be opened if we go this route (especially around selectors for alerts). Will revisit.

PR #254, opened 2023-12-15
chore(charts/app): remove some bad alerts

diranged:
This alarm is pretty noisy and fires off in situations that aren't entirely obvious to engineers. I want to  remove this alarm. Let me explain why.

**Intention**

The motivation of the alarm is to tell an engineer that they aren't running the number of pods that they expect. For example, they're running 5 pods, when 10  are expected. The alarm also tries to be intelligent and filter out times where the `spec.replicas` count has changed and the pods just haven't yet caught up.

**Actual Behavior**

Any time the cluster changes (new nodes, pods evicted, etc) - this rule begins to evaluate as true. Our clusters are extremely dynamic, so workloads are constantly moving around. The larger the workload, the more likely it is that the replica count will not exactly match. For example, if a service has 500 pods spread across 100 hosts ... the odds that at any given time a few pods are not running/not-ready is extremely high.

**Plan?**

I have a plan for a better alarm that will be deployed not at the per-app level, but at the cluster level. I'll open an internal PR for that.


charts/rollout-app/templates/prometheusrules.yaml
```diff
@@ -163,59 +163,6 @@ spec:
   - name: {{ .Release.Namespace }}.{{ .Release.Name }}.{{ .Chart.Name }}.DeploymentRules
     rules:
 
-    {{- with .Values.prometheusRules.DeploymentGenerationMismatch }}
```
diranged, line 166: This alarm never made sense here. These are `Rollout` objects, the `kube_deployment...` metrics simply do not exist for them!

saansh45 (approved):

schahal (approved):
s

PR #256, opened 2024-01-10
feat(charts/nd-common): Adding support for istio sidecar resource override

scohen-nd:
**Motivation**
Due to an incident, we found out that once one of the values is set through annotation. It overides the other values which is what led to the incident.
This behaviour is not documents nor it makes sense in any ways.
**Description**
Once a value is set to any of the istio.(proxyCPU, proxyCPULimit, proxyMemory, proxyMemoryLimit), all the 4 annotations will get created.
Fallbacking to the current implemented default resources for any annotation that is not specifically set:
    Limits:
      cpu:                2
      memory:             1Gi
    Requests:
      cpu:                100m
      memory:             128Mi

The PR includes:
- Adding support for istio sidecar resource override in nd-common istio-annotations.
- Updating dependant charts to use the new nd-common.

### DIFF
--- a/charts/nd-common/Chart.yaml
+++ b/charts/nd-common/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: nd-common
 description: A helper chart used by most of our other charts
 type: library
-version: 0.2.0
+version: 0.2.1
 appVersion: latest
--- a/charts/nd-common/Chart.yaml
+++ b/charts/nd-common/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: nd-common
 description: A helper chart used by most of our other charts
 type: library
-version: 0.2.1
+version: 0.3.0
 appVersion: latest
--- a/charts/nd-common/templates/_istio.tpl
+++ b/charts/nd-common/templates/_istio.tpl
@@ -79,7 +79,7 @@ proxy.istio.io/overrides: >-
   } 
 {{- end }}
 
-{{/*
+{{- /*
 
 Prepare the set of resource annotations for the istio-sidecar.
 Motivation: Due to an incident, we found out that once one of the values is set through annotation. It overides the other values which is what led to the incident.
@@ -100,7 +100,7 @@ Usage: To setup the values follow the regular k8s resources structure under the 
        memory: 10Mi
      limits:
        memory: 60Mi
-*/}}
+*/ -}}
 {{- with .Values.istio.resources }}
 sidecar.istio.io/proxyCPU: {{ .requests.cpu | default "100m" | quote }}
 sidecar.istio.io/proxyCPULimit: {{ .limits.cpu | default "2" | quote }}
--- a/charts/nd-common/templates/_istio.tpl
+++ b/charts/nd-common/templates/_istio.tpl
@@ -82,20 +82,30 @@ proxy.istio.io/overrides: >-
 {{/*
 
 Prepare the set of resource annotations for the istio-sidecar.
-Fallbacking to the current implemented default resources:
-    Limits:
-      cpu:                2
-      memory:             1Gi
-    Requests:
-      cpu:                100m
-      memory:             128Mi
-
+Motivation: Due to an incident, we found out that once one of the values is set through annotation. It overides the other values which is what led to the incident.
+            This behaviour is not documents nor it makes sense in any ways.
+Description: Once a value is set to any of the istio.(proxyCPU, proxyCPULimit, proxyMemory, proxyMemoryLimit), all the 4 annotations will get created.
+             Fallbacking to the current implemented default resources for any annotation that is not specifically set:
+                Limits:
+                  cpu:                2
+                  memory:             1Gi
+                Requests:
+                  cpu:                100m
+                  memory:             128Mi
+Usage: To setup the values follow the regular k8s resources structure under the istio key, like the bellow example:
+  istio:
+   resources:
+     requests:
+       cpu: 10m
+       memory: 10Mi
+     limits:
+       memory: 60Mi
 */}}
-{{- if or (or .Values.istio.proxyCPU .Values.istio.proxyMemory) (or .Values.istio.proxyCPULimit .Values.istio.proxyMemoryLimit) }}
-sidecar.istio.io/proxyCPU: {{ .Values.istio.proxyCPU | default "100m" | quote }}
-sidecar.istio.io/proxyCPULimit: {{ .Values.istio.proxyCPULimit | default "2" | quote }}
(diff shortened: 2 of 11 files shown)


LaikaN57 (approved):

charts/nd-common/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: nd-common
 description: A helper chart used by most of our other charts
 type: library
-version: 0.2.0
+version: 0.2.1
```
diranged, line 5: Since you're adding a new feature - this should probably be `0.3.0`

charts/nd-common/templates/_istio.tpl
```diff
@@ -79,6 +79,25 @@ proxy.istio.io/overrides: >-
   } 
 {{- end }}
 
+{{/*
```
diranged, line 82: Nit. We make careful use of `{{- ` syntax to make sure we don't add newlines to the rendered output when they aren't necessary. For example, this code currently makes this diff:

```diff
 metadata:
   name: simple-app
   labels:
-    helm.sh/chart: simple-app-1.6.0
+    helm.sh/chart: simple-app-1.6.1
     app.kubernetes.io/version: "latest"
     app.kubernetes.io/managed-by: Helm
     tags.datadoghq.com/service: "simple-app"
@@ -242,6 +242,8 @@
               }
             ]
           }
+        
+        
         ad.datadoghq.com/simple-app.check_names: '["prometheus"]'
         ad.datadoghq.com/simple-app.init_configs: '[{}]'
         ad.datadoghq.com/simple-app.instances: |-
@@ -261,7 +263,7 @@
             }
           ]
```

Notice the new-lines? If you change this to `{{- /*` then the diff is:

```diff
 metadata:
   name: simple-app
   labels:
-    helm.sh/chart: simple-app-1.6.0
+    helm.sh/chart: simple-app-1.6.1
     app.kubernetes.io/version: "latest"
     app.kubernetes.io/managed-by: Helm
     tags.datadoghq.com/service: "simple-app"
@@ -261,7 +261,7 @@
             }
           ]
       labels:
-        helm.sh/chart: simple-app-1.6.0
+        helm.sh/chart: simple-app-1.6.1
         app.kubernetes.io/version: "latest"
         app.kubernetes.io/managed-by: Helm
         tags.datadoghq.com/service: "simple-app"
```

charts/nd-common/templates/_istio.tpl
```diff
@@ -79,6 +79,25 @@ proxy.istio.io/overrides: >-
...
+Fallbacking to the current implemented default resources:
+    Limits:
+      cpu:                2
+      memory:             1Gi
+    Requests:
+      cpu:                100m
+      memory:             128Mi
+
+*/}}
+{{- if or (or .Values.istio.proxyCPU .Values.istio.proxyMemory) (or .Values.istio.proxyCPULimit .Values.istio.proxyMemoryLimit) }}
```
diranged, line 94: I don't love that we are copy and pasting the defaults here. An alternative approach is to use `{{- with ...` syntax for each setting. Additionally, I'd love to use standard Kubernetes "resources" syntax. Eg:

```yaml
myApp:
  istio:
    resources:
      requests:
        cpu: 2
        memory: 1Gi
```

Then the template can look something like this:

```
{{- with .Values.istio }}
{{- with .resources }}

{{- with .requests }}
{{- with .cpu }}
sidecar.istio.io/proxyCPU: {{ . | quote }}
{{- end }}
{{- with .memory }}
sidecar.istio.io/proxyMemory: {{ . | quote }}
{{- end }}
{{- end }}

{{- with .limits }}
{{- with .cpu }}
sidecar.istio.io/proxyCPULimit: {{ . | quote }}
{{- end }}
{{- with .memory }}
sidecar.istio.io/proxyMemoryLimit: {{ . | quote }}
{{- end }}
{{- end }}

{{- end }}
{{- end }}
```

## Example - 3 out of 4 settings set:
```diff
+ istio:
+   resources:
+     requests:
+       cpu: 10m
+       memory: 10Mi
+     limits:
+       memory: 60Mi
```

```diff
+    helm.sh/chart: simple-app-1.6.1
     app.kubernetes.io/version: "latest"
     app.kubernetes.io/managed-by: Helm
     tags.datadoghq.com/service: "simple-app"
@@ -242,6 +242,9 @@
               }
             ]
           }
+        sidecar.istio.io/proxyCPU: "10m"
+        sidecar.istio.io/proxyMemory: "10Mi"
+        sidecar.istio.io/proxyMemoryLimit: "60Mi"
         ad.datadoghq.com/simple-app.check_names: '["prometheus"]'
         ad.datadoghq.com/simple-app.init_configs: '[{}]'
         ad.datadoghq.com/simple-app.instances: |-
@@ -261,7 +264,7 @@
```

## Example: Only limits set
```diff
+ istio:
+   resources:
+     requests:
+     limits:
+       memory: 60Mi
```

```diff
   labels:
-    helm.sh/chart: simple-app-1.6.0
+    helm.sh/chart: simple-app-1.6.1
     app.kubernetes.io/version: "latest"
     app.kubernetes.io/managed-by: Helm
     tags.datadoghq.com/service: "simple-app"
@@ -242,6 +242,8 @@
               }
             ]
           }
+        sidecar.istio.io/proxyCPULimit: "400m"
+        sidecar.istio.io/proxyMemoryLimit: "60Mi"
         ad.datadoghq.com/simple-app.check_names: '["prometheus"]'
         ad.datadoghq.com/simple-app.init_configs: '[{}]'
         ad.datadoghq.com/simple-app.instances: |-
@@ -261,7 +263,7 @@
             }
```

scohen-nd: I'm OK with changing the format to match the one from k8s, the with thought will not work.
The reason is that it won't generate the annotation if the value is not set and that's necessary to avoid the issue we had.

diranged: Hmm. Are you saying that we need _all four annotations_ or they won't be used at all? I am under the impression that each annotation can be used individually.

scohen-nd: Yes, the issue was that just changing the cpu requests, cause the memory limits to be set to the memory requests - which cause the sidecar to OOM :-(

scohen-nd: If I set the default values in the value file, I could use your with logic though, would you prefer that?

scohen-nd: You can also look at the current change, which is another approach while leaving the values files empty if you prefer.

scohen-nd: For now I'll leave it like that, we can always change it(the end result is the same).

PR #272, opened 2024-02-16
chore(simple-app): remove uppercase transform of secret engine keys

schahal:
After some private feedback (and I tend to agree), we want to _remove_ the `uppercase` transformation on secret property keys.

As the rest of the engines (`plaintext` and `kms`) are in abundant use, changing that would be BREAKING. 

I feel we can switch the `sealed` engine template only, as adoption of that is nil (and IMO a breaking change later for the rest of the secret engines).

I think the advantage of letting user explicitly set the key name (rather than doing a "behind-the-scenes" transformation) outweigh the cons (less friction finding/debugging secrets, some 3rd-party charts may require user to have a secret with a specifically styled key, etc)

### DIFF
--- a/charts/daemonset-app/templates/sealedsecret.yaml
+++ b/charts/daemonset-app/templates/sealedsecret.yaml
@@ -13,7 +13,7 @@ metadata:
 spec:
   encryptedData:
     {{- range $key, $value := .Values.secrets }}
-    {{ $key | upper }}: {{ $value | quote }}
+    {{ $key }}: {{ $value | quote }}
     {{- end }}
   template:
     metadata:
--- a/charts/daemonset-app/Chart.yaml
+++ b/charts/daemonset-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: daemonset-app
 description: Default DaemonSet Helm Chart
 type: application
-version: 0.15.3
+version: 0.15.4
 appVersion: latest
 maintainers:
   - name: diranged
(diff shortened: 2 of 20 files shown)


charts/daemonset-app/templates/sealedsecret.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
 spec:
   encryptedData:
     {{- range $key, $value := .Values.secrets }}
-    {{ $key | upper }}: {{ $value | quote }}
+    {{ $key }}: {{ $value | quote }}
```
diranged, line 16: So .. I'm a bit confused. Why do we think this is a breaking change exactly? Are people using lower-cased environment variables?

schahal: It's changing the behavior of the underlying template (people _may_ have been relying on this behavior).

That's why I thought not doing the same change for the existing `secrets.yaml` and  `kmssecrets.yaml`... but I'd like the added consistency and maybe we can make it part of this change (and do a maj version bump)?

diranged: Ok. I don't feel too strongly - but I feel like the underlying  chart behavior should change and we should jsut consider that a minor version bump. It's reasonable that environment variables should be upper-cased..

schahal: Great! My heart wanted to make it this behavior for everyone :) (but didn't want to step on any previous decisions that were made regarding that). Updated (and glad we did)!

diranged (approved):

PR #273, opened 2024-02-21
chore(stateful-app): Explicitly set the default UpdateStrategy

schahal:
Can I get thoughts on explicitly setting the default?

`RollingUpdate` becomes the default if nothing (`null`) is set, anyways.

I see some advantages to this:
- Users don't have to guess on the DeploymentStrategy anymore
- They can explicitly see where to update partition if they want to, for example, do a canary rollout
- [Highly unlikely] If default were to change in a future k8s release (realistically this will only even be plausible in a majore release, not 1.x), at least apps subcharting this will stay consistent.

(but I may be missing some historical context on why we want null)

### DIFF
--- a/charts/stateful-app/Chart.yaml
+++ b/charts/stateful-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: stateful-app
 description: Default StatefulSet Helm Chart
 type: application
-version: 0.16.4
+version: 0.16.5
 appVersion: latest
 maintainers:
   - name: diranged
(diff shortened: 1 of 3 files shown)


diranged:
My general feeling is that I don't like copying and pasting defaults ... so I look to this and ask, "What problem are we solving?"

schahal:
Closing out this PR (for context, we chatted offline.. we'll stick to the pattern not copy-pasting defaults in).

In general, if a user wants to fiddle with deployment strategies, they are already aware of what they'd want to change and can do so in their own "values.yaml" overlay (the problem to solve of seeing which default deployment strategy used isn't _**that**_ great to change the general rule above)

PR #278, opened 2024-03-22
[Flink operator] Rename pod monitor label from cluster to flink-cluster

annagnd:
Changing pod monitor label from `cluster` to `flink-cluster` for flonk-operator, per [this](https://nextdoor.slack.com/archives/CM6EWSYH4/p1710968293931579) discussion.

### DIFF
--- a/charts/flink-operator/Chart.yaml
+++ b/charts/flink-operator/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Intalls the Spotify-version of the Flink-on-K8S-Operator
 name: flink-operator
-version: 0.0.1
+version: 0.1.0
 appVersion: v0.3.0
 sources:
   - https://github.com/spotify/flink-on-k8s-operator/releases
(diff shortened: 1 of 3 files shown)


charts/flink-operator/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 description: Intalls the Spotify-version of the Flink-on-K8S-Operator
 name: flink-operator
-version: 0.0.1
```
nsreenivasamurthy-nextdoor, line 4: This is a patch version change (Semantic version is usually of this format: _<major>.<minor>.<patch>_). Changing the middle number would be better for this PR, unless this is an automatic system generated change.

nsreenivasamurthy-nextdoor (approved):

charts/flink-operator/Chart.yaml
nsreenivasamurthy-nextdoor: Format: MAJOR.MINOR.PATCH

diranged (approved):

PR #285, opened 2024-04-15
feat(prometheus-alerts): Add alerts to check that selectors actually select valid data

LaikaN57:

I tried to choose metrics for each selector that were generic enough and did not rely on a specific metric that may not be gathered. Most types had an associated `info` or `label` metric which are good candidates for this use-case.

Proof:

image

Here we can see that query A (bad selector) does not return any metrics, therefore gets zero filled, and we trigger an alert. Contrarily, we see that query B (good selector) does have series associated and therefore we do not alert.

### DIFF
--- a/charts/istio-alerts/templates/service-prometheusrule.yaml
+++ b/charts/istio-alerts/templates/service-prometheusrule.yaml
@@ -56,7 +56,7 @@ spec:
       annotations:
         summary: >-
           {{`{{$labels.destination_service_name}}`}} avg request latencies are above {{ .threshold }}s!
-        runbook_url: {{ $values.defaults.runbookUrl}} #HighRequestLatency
+        runbook_url: {{ $values.defaults.runbookUrl}}#HighRequestLatency
         description: >-
           Average request latency of {{`{{ $value | humanizePercentage }}`}} is above threshold ({{ .threshold }}s)
           in namespace {{`{{ $labels.namespace }}`}} for pod {{`{{ $labels.pod }}`}} (container: {{`{{ $labels.container }}`}}).
@@ -85,5 +85,37 @@ spec:
         {{ toYaml . | nindent 8}}
         {{- end }}
     {{- end }}
-  {{- end }}
+    {{- end }}
+
+  - name: {{ .Release.Name }}.{{ .Release.Namespace }}.{{ .Chart.Name }}.alertRulesSelectorsValidity
+    rules:
+    {{- with .Values.serviceRules.destinationServiceSelectorValidity }}
+    - alert: DestinationServiceSelectorValidity
+      annotations:
+        summary: >-
+          DestinationServiceSelector for istio-alerts is invalid
+        runbook_url: {{ $values.defaults.runbookUrl}}#Alert-Rules-Selectors-Validity
+        description: >-
+          The DaemonsetSelector used for daemonset level alerts did not return any data.
+          Please check the DaemonsetSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your daemonsets so that you
+          will be alerted for daemonset issues. The current selector is
+          `{destination_service_namespace="{{ $release.Namespace }}", destination_service_name="{{ $.Values.serviceRules.destinationServiceName }}"}`.
+      expr: |-
+        (
+          count(
+            istio_request_duration_milliseconds_bucket{
+              destination_service_namespace="{{ $release.Namespace }}",
+              destination_service_name=~"{{ $.Values.serviceRules.destinationServiceName }}"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: {{ .for }}
+      labels:
+        severity: {{ .severity }}
+        namespace: {{ $release.Namespace }}
+        {{- with $values.defaults.additionalRuleLabels}}
+        {{ toYaml . | nindent 8}}
+        {{- end }}
+    {{- end }}
 {{- end }}
--- a/charts/istio-alerts/values.yaml
+++ b/charts/istio-alerts/values.yaml
@@ -117,3 +117,7 @@ serviceRules:
 
     # -- How long to evaluate the latency of services.
     for: 15m
+  
+  destinationServiceSelectorValidity:
+    sevarity: warning
+    for: 1h
--- a/charts/istio-alerts/values.yaml
+++ b/charts/istio-alerts/values.yaml
@@ -118,6 +118,13 @@ serviceRules:
     # -- How long to evaluate the latency of services.
     for: 15m
 
+  # -- Does a basic lookup using the defined selectors to see if we can see any
+  # info for a given selector. This is the "watcher for the watcher". If we get
+  # alerted by this, we likely have a bad selector and our alerts are not going
+  # to ever fire.
   destinationServiceSelectorValidity:
+    # -- Severity of the monitor
     sevarity: warning
+
+    # -- How long to evaluate.
     for: 1h
--- a/charts/istio-alerts/values.yaml
+++ b/charts/istio-alerts/values.yaml
@@ -117,7 +117,7 @@ serviceRules:
 
     # -- How long to evaluate the latency of services.
     for: 15m
-  
+
   destinationServiceSelectorValidity:
     sevarity: warning
     for: 1h
--- a/charts/istio-alerts/values.yaml
+++ b/charts/istio-alerts/values.yaml
@@ -124,7 +124,7 @@ serviceRules:
   # to ever fire.
   destinationServiceSelectorValidity:
     # -- Severity of the monitor
-    sevarity: warning
+    severity: warning
 
     # -- How long to evaluate.
     for: 1h
--- a/charts/prometheus-alerts/templates/containers-prometheusrule.yaml
+++ b/charts/prometheus-alerts/templates/containers-prometheusrule.yaml
@@ -480,4 +480,154 @@ spec:
         {{- end }}
     {{- end }}
 
+  - name: {{ .Release.Name }}.{{ .Release.Namespace }}.alertRulesSelectorsValidity
+    rules:
+    {{ with .Values.containerRules.AlertRulesSelectorsValidity -}}
+    - alert: DaemonsetSelectorValidity
+      annotations:
+        summary: DaemonsetSelector for prometheus-alerts is invalid
+        runbook_url: {{ $values.defaults.runbookUrl }}#Alert-Rules-Selectors-Validity
+        description: >-
+          The DaemonsetSelector used for daemonset level alerts did not return any data.
+          Please check the DaemonsetSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your daemonsets so that you
+          will be alerted for daemonset issues. The current selector is
+          `{daemonsetSelector="{{ $daemonsetSelector }}"}`.
+      expr: |-
+        (
+          count(
+            kube_daemonset_labels{
+              {{ $daemonsetSelector }}
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: {{ .for }}
+      labels:
+        severity: {{ .severity }}
+        {{- if $values.defaults.additionalRuleLabels }}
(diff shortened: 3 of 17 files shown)


LaikaN57:
```diff
--- orig	2024-04-15 17:16:43
+++ new	2024-04-15 17:15:45
@@ -53,7 +53,7 @@
     nextdoor.com/chart: prometheus-rules
     nextdoor.com/source: https://github.com/Nextdoor/k8s-charts
   labels:
-    helm.sh/chart: prometheus-alerts-1.4.0
+    helm.sh/chart: prometheus-alerts-1.5.0
     app.kubernetes.io/version: "0.0.1"
     app.kubernetes.io/managed-by: Helm

@@ -431,5 +431,135 @@
           ==
         kube_horizontalpodautoscaler_spec_max_replicas{job="kube-state-metrics", horizontalpodautoscaler=~"prometheus-alerts-.*", namespace="monolith"}
       for: 15m
+      labels:
+        severity: warning
+
+  - name: prometheus-alerts.monolith.alertRulesSelectorsValidity
+    rules:
+    - alert: DaemonsetSelectorValidity
+      annotations:
+        summary: DaemonsetSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The DaemonsetSelector used for daemonset level alerts did not return any data.
+          Please check the DaemonsetSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your daemonsets so that you
+          will be alerted for daemonset issues. The current selector is
+          `{daemonsetSelector="job="kube-state-metrics", daemonset=~"prometheus-alerts-.*", namespace="monolith""}`.
+      expr: |-
+        (
+          count(
+            kube_daemonset_labels{
+              job="kube-state-metrics", daemonset=~"prometheus-alerts-.*", namespace="monolith"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
+      labels:
+        severity: warning
+    - alert: DeploymentSelectorValidity
+      annotations:
+        summary: DeploymentSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The DeploymentSelector used for deployment level alerts did not return any data.
+          Please check the DeploymentSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your deployments so that you
+          will be alerted for deployment issues. The current selector is
+          `{deploymentSelector="job="kube-state-metrics", kube_deployment=~"prometheus-alerts-.*", namespace="monolith""}`.
+      expr: |-
+        (
+          count(
+            kube_deployment_labels{
+              job="kube-state-metrics", kube_deployment=~"prometheus-alerts-.*", namespace="monolith"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
+      labels:
+        severity: warning
+    - alert: PodSelectorValidity
+      annotations:
+        summary: PodSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The PodSelector used for pod level alerts did not return any data.
+          Please check the PodSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your pods so that you
+          will be alerted for pod issues. The current selector is
+          `{podSelector="pod=~"prometheus-alerts-.*"", namespaceSelector="namespace="monolith""}`.
+      expr: |-
+        (
+          count(
+            kube_pod_info{
+              pod=~"prometheus-alerts-.*",
+              namespace="monolith"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
+      labels:
+        severity: warning
+    - alert: JobSelectorValidity
+      annotations:
+        summary: JobSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The JobSelector used for job level alerts did not return any data.
+          Please check the JobSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your jobs so that you
+          will be alerted for job issues. The current selector is
+          `{jobSelector="job="kube-state-metrics", job_name=~"prometheus-alerts-.*", namespace="monolith""}`.
+      expr: |-
+        (
+          count(
+            kube_job_info{
+              job="kube-state-metrics", job_name=~"prometheus-alerts-.*", namespace="monolith"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
+      labels:
+        severity: warning
+    - alert: HpaSelectorValidity
+      annotations:
+        summary: HpaSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The HpaSelector used for hpa level alerts did not return any data.
+          Please check the HpaSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your hpas so that you
+          will be alerted for hpa issues. The current selector is
+          `{hpaSelector="job="kube-state-metrics", horizontalpodautoscaler=~"prometheus-alerts-.*", namespace="monolith""}`.
+      expr: |-
+        (
+          count(
+            kube_horizontalpodautoscaler_info{
+              job="kube-state-metrics", horizontalpodautoscaler=~"prometheus-alerts-.*", namespace="monolith"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
+      labels:
+        severity: warning
+    - alert: StatefulsetSelectorValidity
+      annotations:
+        summary: StatefulsetSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The StatefulsetSelector used for statefulset level alerts did not return any data.
+          Please check the StatefulsetSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your statefulsets so that you
+          will be alerted for statefulset issues. The current selector is
+          `{statefulsetSelector="job="kube-state-metrics", statefulset=~"prometheus-alerts-.*", namespace="monolith""}`.
+      expr: |-
+        (
+          count(
+            kube_statefulset_created{
+              job="kube-state-metrics", statefulset=~"prometheus-alerts-.*", namespace="monolith"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
       labels:
         severity: warning
```

charts/istio-alerts/values.yaml
```diff
@@ -117,3 +117,14 @@ serviceRules:
...
     # -- How long to evaluate the latency of services.
     for: 15m
+
+  # -- Does a basic lookup using the defined selectors to see if we can see any
+  # info for a given selector. This is the "watcher for the watcher". If we get
+  # alerted by this, we likely have a bad selector and our alerts are not going
+  # to ever fire.
+  destinationServiceSelectorValidity:
+    # -- Severity of the monitor
+    sevarity: warning
```
diranged, line 127: Typo. `severity`.

charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -85,5 +85,37 @@ spec:
...
+  - name: {{ .Release.Name }}.{{ .Release.Namespace }}.{{ .Chart.Name }}.alertRulesSelectorsValidity
+    rules:
+    {{- with .Values.serviceRules.destinationServiceSelectorValidity }}
+    - alert: DestinationServiceSelectorValidity
+      annotations:
+        summary: >-
+          DestinationServiceSelector for istio-alerts is invalid
+        runbook_url: {{ $values.defaults.runbookUrl}}#Alert-Rules-Selectors-Validity
+        description: >-
+          The DaemonsetSelector used for daemonset level alerts did not return any data.
```
diranged, line 99: Copy-paste error... this message should be accurate to the alarm, it's about istio..

charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -85,5 +85,37 @@ spec:
...
+          Please check the DaemonsetSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your daemonsets so that you
+          will be alerted for daemonset issues. The current selector is
+          `{destination_service_namespace="{{ $release.Namespace }}", destination_service_name="{{ $.Values.serviceRules.destinationServiceName }}"}`.
```
diranged, line 103: So - I think we need to create functions now that define the selectors. We should create a function at the top of the file like this:

```
{{- define "istioAlerts.selectors" -}}
printf "destination_service_namespace=\"%s\", "destination_service_name=\"%s\"" $.Release.Namespace $.Values.serviceRules.destinationServiceName }}
{{- end }}
```

Then in each of the alerts, we can use that function... ie:

```promql
        histogram_quantile(
          {{ .percentile }},
          sum(irate(
            istio_request_duration_milliseconds_bucket{
              {{ include "istioAlerts.selectors" $ }}
            }[5m]
          )) by (
            destination_service_name,
            reporter,
            source_canonical_service,
            le
          )
        )
        / 1000
        > {{ .threshold }}
```

(we should do this in all of the alerts)

charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -85,5 +85,37 @@ spec:
...
+      expr: |-
+        (
```
diranged, line 105: I think I see what you're doing here ...  but I see a risk. If there is no traffic to the service for example, when its being launched), there will be no metrics that match this query. I propose that instead we do this:

```
count(kube_service_info{namespace={{ $.Release.Namespace }}, name={{ $.Release.Namespace $.Values.serviceRules }})
```

Sadly - I know that this means we aren't using the exact selectors (like I propose above) ... but this allows us to verify that the Service actually exists by that name. Ya know?

Thoughts?

charts/prometheus-alerts/templates/containers-prometheusrule.yaml
```diff
@@ -480,4 +480,154 @@ spec:
         {{- end }}
     {{- end }}
 
+  - name: {{ .Release.Name }}.{{ .Release.Namespace }}.alertRulesSelectorsValidity
```
diranged, line 483: Unless I am misreading - the problem I see here is that this chart is launched in tons of places and it creates rules that may never be used .. ie, it creates DaemonSet rules even if the namespace has no DaemonSets. That's not _great_ .. but it previously didn't cause any problems. Now we're talking about creating an alarm that I think will fire off for almost everyone when it gets created because they are mostly not using all of these resources.

Eg.. I just ran this with some fake selectors:

image

For the `simple-app`, `daemonset-app`, `rollout-app` charts, we could go down the route of adding in options that simply turn off the unrelated alarms (ie... add in `.Values.containerRules.enableDaemonsetRules` for example).

For _outside_ callers of this chart, we'll need a big noisy readme breaking change changelog that tells users to add these variables in or they'll get paged.

Thoughts?

charts/istio-alerts/values.yaml
LaikaN57, line 127:
```suggestion
    severity: warning
```

LaikaN57:
@diranged I am actually going to split this PR into smaller chunks so we can modify `prometheus-alerts`/`istio-alerts` charts individually as I think this PR is going to get too conflated.

charts/istio-alerts/templates/service-prometheusrule.yaml
LaikaN57: resolved in new PR

charts/istio-alerts/templates/service-prometheusrule.yaml
LaikaN57: I will do this in a separate small PR. skipping for now

PR #287, opened 2024-04-17
feat(istio-alerts): Add service selector validity alert

LaikaN57:
> [!WARNING]
> Reminder for @LaikaN57 - Update other charts in this repo after this PR is merged.

Split from https://github.com/Nextdoor/k8s-charts/pull/285

Snippet from the other PR about using `kube_service_info` as our test metric.

> I think I see what you're doing here ...  but I see a risk. If there is no traffic to the service for example, when its being launched), there will be no metrics that match this query. I propose that instead we do this:
>
> ```
> count(kube_service_info{namespace={{ $.Release.Namespace }}, name={{ $.Release.Namespace $.Values.serviceRules }})
> ```
>
> Sadly - I know that this means we aren't using the exact selectors (like I propose above) ... but this allows us to verify that the Service actually exists by that name. Ya know?

### DIFF
--- a/charts/istio-alerts/README.md.gotmpl
+++ b/charts/istio-alerts/README.md.gotmpl
@@ -1,0 +1,21 @@
+{{ template "chart.header" . }}
+
+{{ template "chart.versionBadge" .  }}{{ template "chart.typeBadge" .  }}{{ template "chart.appVersionBadge" .  }}
+
+{{ template "chart.description" . }}
+
+{{ template "chart.maintainersSection" . }}
+
+## Upgrade Notes
+
+### 0.2.x -> 0.3.x
+
+**BREAKING: TBD**
+
+TBD
+
+{{ template "chart.requirementsSection" . }}
+
+{{ template "chart.valuesSection" . }}
+
+{{ template "helm-docs.versionFooter" . }}
--- a/charts/istio-alerts/README.md.gotmpl
+++ b/charts/istio-alerts/README.md.gotmpl
@@ -10,9 +10,12 @@
 
 ### 0.2.x -> 0.3.x
 
-**BREAKING: TBD**
+**BREAKING: The DestinationServiceSelectorValidity alert rule requires kube-state-metrics.**
 
-TBD
+An alert was introduced in 0.3.x that requires kube-state-metrics to be installed in the cluster. If
+you do not have kube-state-metrics installed, you will need to disable the alert by setting
+`serviceRules.destinationServiceSelectorValidity.enabled` to `false`. This alert is used to detect
+if the destinationServiceSelector is actually selecting series for a service that exists.
 
 {{ template "chart.requirementsSection" . }}
 
--- a/charts/istio-alerts/templates/service-prometheusrule.yaml
+++ b/charts/istio-alerts/templates/service-prometheusrule.yaml
@@ -22,7 +22,6 @@ spec:
         description: >-
           High rate of 5xx responses from the {{`{{$labels.destination_service_name}}`}} VirtualService
           in namespace {{`{{$labels.namespace}}`}}.
-
       expr: >
         sum by (destination_service_name, reporter, source_workload) (
           istio_requests:increase5m{
@@ -48,15 +47,13 @@ spec:
         {{- end }}
     {{- end }}
     {{- end }}
-
-     # Adding latency alarm to trigger when request latency is above > threshold
     {{- with .Values.serviceRules.highRequestLatency }}
     {{- if .enabled }}
     - alert: HighRequestLatency
       annotations:
         summary: >-
           {{`{{$labels.destination_service_name}}`}} avg request latencies are above {{ .threshold }}s!
-        runbook_url: {{ $values.defaults.runbookUrl}} #HighRequestLatency
+        runbook_url: {{ $values.defaults.runbookUrl}}#HighRequestLatency
         description: >-
           Average request latency of {{`{{ $value | humanizePercentage }}`}} is above threshold ({{ .threshold }}s)
           in namespace {{`{{ $labels.namespace }}`}} for pod {{`{{ $labels.pod }}`}} (container: {{`{{ $labels.container }}`}}).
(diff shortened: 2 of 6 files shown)


charts/istio-alerts/README.md.gotmpl
```diff
@@ -0,0 +1,21 @@
...
+
+{{ template "chart.maintainersSection" . }}
+
+## Upgrade Notes
+
+### 0.2.x -> 0.3.x
+
+**BREAKING: TBD**
+
+TBD
```
LaikaN57, line 15: WIP

charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -85,5 +82,36 @@ spec:
...
+      annotations:
+        summary: >-
+          DestinationServiceSelector for istio-alerts is invalid
+        runbook_url: {{ $values.defaults.runbookUrl}}#Alert-Rules-Selectors-Validity
+        description: >-
+          The DestinationServiceSelector used for service level alerts did not return any data.
+          Please check the DestinationServiceSelector applied in your istio-alerts chart
+          is correct to ensure you are properly selecting your service so that you
+          will be alerted for service issues. The current selector is
+          `{destination_service_namespace="{{ $release.Namespace }}", destination_service_name="{{ $.Values.serviceRules.destinationServiceName }}"}`.
```
LaikaN57, line 98: WIP

charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -85,5 +82,36 @@ spec:
...
+      expr: >-
+        (
+          count(
+            kube_service_info{
+              namespace={{ $.Release.Namespace }},
+              name={{ $.Values.serviceRules.destinationServiceName }}
```
LaikaN57, line 104: WIP

charts/istio-alerts/templates/service-prometheusrule.yaml
diranged, line 98: Can we replace these hard-coded lists of the selectors with a variable - that way you can just refer to `$istioSelector`? We started doing that at https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/templates/containers-prometheusrule.yaml#L2-L8 and https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/templates/_helpers.tpl#L13-L59 and it makes the code more readable and easier to modify later on.

charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -85,5 +82,36 @@ spec:
...
+          The DestinationServiceSelector used for service level alerts did not return any data.
+          Please check the DestinationServiceSelector applied in your istio-alerts chart
+          is correct to ensure you are properly selecting your service so that you
+          will be alerted for service issues. The current selector is
+          `{destination_service_namespace="{{ $release.Namespace }}", destination_service_name="{{ $.Values.serviceRules.destinationServiceName }}"}`.
+      expr: >-
+        (
+          count(
+            kube_service_info{
+              namespace={{ $.Release.Namespace }},
```
diranged, line 103: Pretty sure these need to be quoted anyways...

charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -85,5 +82,36 @@ spec:
...
+              name={{ $.Values.serviceRules.destinationServiceName }}
+            }
+          ) or on() vector(0)
```
diranged, line 106: So this part of the alert is a little tricky ... if you're missing the metric, it shows `0`:
image

However, if we typed the metric name wrong, it also reports `0`.. (or as you pointed out, if you are missing kube-state-metrics):
image

Perhaps we just want to add a second rule in here (thats under the same condition) that verifies that kube-state-metrics is working too and if its not, sends a separate alarm with useful info to tell us that we broke something related to the kube-state-metrics?

```
(count(up{job="kube-state-metrics"}) or on() vector(0) == 0)
```

or

```
(count(kube_service_info{}) or on() vector(0) == 0)
```

charts/istio-alerts/values.yaml
```diff
@@ -117,3 +118,17 @@ serviceRules:
...
+  # -- Does a basic lookup using the defined selectors to see if we can see any
+  # info for a given selector. This is the "watcher for the watcher". If we get
+  # alerted by this, we likely have a bad selector and our alerts are not going
+  # to ever fire.
+  destinationServiceSelectorValidity:
+    # -- Whether to enable the monitor on the selector for the VirtualService.
+    enabled: true
+
+    # -- Severity of the monitor
+    severity: warning
```
diranged, line 131: Are we going to set this to critical later - or leave it as warning? if we're leaving it as warning (and it feels important), let's maybe describe our thinking here?

charts/istio-alerts/templates/service-prometheusrule.yaml
LaikaN57: I actually want to split that into a follow up PR as I do not want to conflate the code change of adding a new metric with changing the code of the existing metrics.

charts/istio-alerts/templates/service-prometheusrule.yaml
LaikaN57: I think in order to avoid scope creep, in this chart we have to assume KSM is working. In the README, KSM is now listed as a requirement and there is a flag to disable this alert if you do not want to install KSM.

charts/istio-alerts/values.yaml
LaikaN57: This was a pattern match to the other alerts. I don't know when/why we are allowing the severity to be set. My gut says that most users are probably not overriding this.

charts/istio-alerts/templates/service-prometheusrule.yaml
LaikaN57: KSM health tracking should be done in a KSM related chart. It would not make sense for all KSM dependant charts to have independant/redundant KSM health alerts in place IMO.

charts/istio-alerts/values.yaml
diranged: So the question to you is - is this a critical alert, or do we consider this a warning? We want the default to be "right" here for most cases. What is your gut?

charts/istio-alerts/templates/service-prometheusrule.yaml
diranged: I'll accept that for now ... I reserve the right to come back to this discussion. :P

diranged (approved):

charts/istio-alerts/values.yaml
LaikaN57: So this is a warning because it is not saying your actual inspected thing is actively breaking. it is just saying you have loss of observability. Depending on the service this could be configured to be a critical problem if you prefer to be really on top of monitoring. This is not a reality for all teams/end-users tho due to prioritization. So I think default to warning and let it be configured for critical is the right path here.

PR #289, opened 2024-04-17
chore(istio-alerts): Update dependant parent charts

LaikaN57:


charts/rollout-app/Chart.yaml
```diff
@@ -2,14 +2,14 @@ apiVersion: v2
 name: rollout-app
 description: Argo Rollout-based Application Helm Chart
 type: application
-version: 0.7.4
+version: 0.7.5
```
diranged, line 5: Given the major change in the `istio-alerts` behavior, we should bump the minor versions here I believe.

LaikaN57: updated

charts/stateful-app/Chart.yaml
```diff
@@ -2,14 +2,14 @@ apiVersion: v2
 name: stateful-app
 description: Default StatefulSet Helm Chart
 type: application
-version: 0.16.4
```
LaikaN57, line 5: This really should have been 1.0.0 already actually I think because this is no longer beta.

charts/simple-app/README.md
```diff
@@ -13,13 +13,18 @@ defaults for you like the Kubernetes [Horizontal Pod Autoscaler][hpa].
 
 ## Upgrade Notes
 
-### 1.2.x -> 1.7.x
+### 1.6.x -> 1.7.x
```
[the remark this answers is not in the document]
LaikaN57: if someone jumps from 1.2 or 1.3 or 1.4 -> 1.7 then they should logically conclude they need to be aware of this note as well.

corydolphin (approved):

charts/simple-app/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-app
 description: Default Microservice Helm Chart
 type: application
-version: 1.7.0
+version: 1.7.1
```
corydolphin, line 5: I don't think you ned to version change for a readme change.

PR #290, opened 2024-04-17
feat(charts/istio-alerts): Part1: Parameterize the set of labels used for grouping 5XX alarms

corydolphin:
**Context**
In https://github.com/Nextdoor/k8s-charts/commit/c8308088b2acdf509ee3c3c5c431534710210fd8 there was a change to the way the service level 5XX alarm worked such that it computes an alarms on the 5XX rate individually for each source service. 

While it is valuable to have an alarm from the "client"'s perspective from reporting, and provide capabilities for there to be an alarm on arbitrary grouping keys, this wasn't the historical behavior, and the alarm has a different meaning. 

**Proposal**
To improve the behavior, I am first moving the definition of these labels into a value so that in a subsequent PR we can change the default behavior to match the intention before https://github.com/Nextdoor/k8s-charts/commit/c8308088b2acdf509ee3c3c5c431534710210fd8 

The intention o this change is to have no impact on the behavior, and just move the definition. See helm template output below

```
➜  istio-alerts git:(main) ✗ diff -bu old.yaml new.yaml
--- old.yaml	2024-04-17 16:51:22
+++ new.yaml	2024-04-17 16:51:09
@@ -49,6 +49,7 @@
   groups:
   - name: observability.istio-alerts.istio-alerts.IstioServiceRules
     rules:
+
     - alert: 5xx-Rate-Too-High
       annotations:
         summary: >-
```

### DIFF
--- a/charts/istio-alerts/templates/service-prometheusrule.yaml
+++ b/charts/istio-alerts/templates/service-prometheusrule.yaml
@@ -17,7 +17,7 @@ spec:
     rules:
     {{- with .Values.serviceRules.http5XXMonitor }}
     {{- if .enabled }}
-    {{ $http5XXMonitorGroupingLabelString := join ", " .monitorGroupingLabels }}
+    {{- $http5XXMonitorGroupingLabelString := join ", " .monitorGroupingLabels }}
     - alert: 5xx-Rate-Too-High
       annotations:
         summary: >-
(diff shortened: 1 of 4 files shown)


charts/istio-alerts/templates/service-prometheusrule.yaml
```diff
@@ -16,6 +17,7 @@ spec:
     rules:
     {{- with .Values.serviceRules.http5XXMonitor }}
     {{- if .enabled }}
+    {{ $http5XXMonitorGroupingLabelString := join ", " .monitorGroupingLabels }}
```
diranged, line 20: Nit: use `{{-` to ensure no additional newline is created.

diranged (commented):
One tiny nit - overall looks great.

diranged (approved):

PR #292, opened 2024-04-18
feat(prometheus-alerts): Add selector validity alerts

LaikaN57:

I tried to choose metrics for each selector that were generic enough and did not rely on a specific metric that may not be gathered. Most types had an associated info or label metric which are good candidates for this use-case.

See https://github.com/Nextdoor/k8s-charts/pull/285


LaikaN57:
TODO: https://github.com/Nextdoor/k8s-charts/pull/285#discussion_r1566695898

charts/prometheus-alerts/README.md
```diff
@@ -20,6 +19,19 @@ those changes in the `charts/simple-app`, `charts/daemonset-app` and
 
 ## Upgrade Notes
 
+### 1.4.x -> 1.5.x
+
+**BREAKING: The AlertSelectorValidity alert rules requires kube-state-metrics.**
+
+An alert was introduced in that requires kube-state-metrics to be installed in the cluster. If
```
diranged, line 26: All of these metrics rely on kube-state-metrics ... this is not a change from the previous behavior.

charts/prometheus-alerts/templates/containers-prometheusrule.yaml
```diff
@@ -480,4 +480,166 @@ spec:
         {{- end }}
     {{- end }}
 
+  - name: {{ .Release.Name }}.{{ .Release.Namespace }}.alertRulesSelectorsValidity
```
diranged, line 483: Ok. This chart maybe needs a little more reworking than this. I think that each of the core types (DaemonSets, Deployments, StatefulSets, Jos, HPAs) should have its own settings section. Each of those sections should be controlled with an on-off flag, and it should turn off ALL the alerts.

Eg:

```yaml
containerRules:
  enabled: true

  pods:
    enabled: true
    PodContainerTerminated:
      severity: warning
      threshold: 0
      over: 10m
      for: 1m
      reasons:
        # - Error  < when a container is evicted gracefully, the "error" state is used.
        - ContainerCannotRun
        - DeadlineExceeded

  deployments:
    enabled: true
    KubeDeploymentGenerationMismatch:
      severity: warning
      for: 15m

  statefulSets:
    enabled: true
    KubeStatefulSetReplicasMismatch:
      severity: warning
      for: 15m
```

Does that make sense? This would obviously be a breaking change - so we probably would add in a safety check that errors out if it finds old values (`.Values.containerRules.PodContainerTerminated`, etc..). Alternatively, we could go as far as implementing a schema file ... but that might be just a pain in the ass.

LaikaN57: yeah, ok i was thinking re-writing the entire chart as well... I like the safety check idea. That was my main concern there with re-writing.

LaikaN57: I think this is going to be a huge change so let me restructure first in 1 PR and then add the new alerts back in a second follow up PR.

LaikaN57: https://github.com/Nextdoor/k8s-charts/pull/295

PR #293, opened 2024-04-18
fix(charts/simple-app): Update istio-alerts

corydolphin:
This propagate the istio-alerts update. It must be done as a separate PR and only merged after due to the use of http for referencing the charts repo


LaikaN57 (commented):
Can you please bump all the charts in the repo that use this, please. Other than that, this looks good to me.

LaikaN57 (approved):

PR #294, opened 2024-04-18
fix(nd-common): Repoint parent charts back to published chart repo

LaikaN57:


diranged (approved):
I think this is good - could we add a followup to add renovatebot here so that it will self-update these charts for us? (in a different PR)

PR #297, opened 2024-04-23
feat(prometheus-alerts): Add selector validity alarms

LaikaN57:
We ran into an issue where unbeknownst to us, our pod selector was set incorrectly and therefore all of our rules deployed via prometheus-alerts did not have any series to evaluate. This lack of series data is silently ignored. Here we are resolving this by implementing a rule which will use the various selectors to see if a basic metric exists or not. If it does not exist then we want to warn the user that none of the other rules will be evaluated.

I tried to choose metrics for each selector that were generic enough and did not rely on a specific metric that may not be gathered. Most types had an associated info or label metric which are good candidates for this use-case.

See https://github.com/Nextdoor/k8s-charts/pull/285

### DIFF
--- a/charts/prometheus-alerts/templates/containers-prometheusrule.yaml
+++ b/charts/prometheus-alerts/templates/containers-prometheusrule.yaml
@@ -200,6 +200,34 @@ spec:
         {{- end }}
     {{- end }}
 
+    {{ with .PodSelectorValidity -}}
+    - alert: PodSelectorValidity
+      annotations:
+        summary: PodSelector for prometheus-alerts is invalid
+        runbook_url: {{ $.Values.defaults.runbookUrl }}#Alert-Rules-Selectors-Validity
+        description: >-
+          The PodSelector used for pod level alerts did not return any data.
+          Please check the PodSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your pods so that you
+          will be alerted for pod issues. The current selector is
+          `{podSelector="{{ $podSelector }}", namespaceSelector="{{ $namespaceSelector }}"}`.
+      expr: |-
+        (
+          count(
+            kube_pod_info{
+              {{ $podSelector }},
+              {{ $namespaceSelector }}
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: {{ .for }}
+      labels:
+        severity: {{ .severity }}
+        {{- if $.Values.defaults.additionalRuleLabels }}
+        {{ toYaml $.Values.defaults.additionalRuleLabels | nindent 8 }}
+        {{- end }}
+    {{- end }}
+
     {{- end }}
     {{- end }}
 
@@ -220,6 +248,33 @@ spec:
         kube_deployment_status_observed_generation{ {{- $deploymentSelector -}} }
           !=
         kube_deployment_metadata_generation{ {{- $deploymentSelector -}} }
+      for: {{ .for }}
+      labels:
+        severity: {{ .severity }}
+        {{- if $.Values.defaults.additionalRuleLabels }}
+        {{ toYaml $.Values.defaults.additionalRuleLabels | nindent 8 }}
+        {{- end }}
+    {{- end }}
+
+    {{ with .DeploymentSelectorValidity -}}
+    - alert: DeploymentSelectorValidity
+      annotations:
+        summary: DeploymentSelector for prometheus-alerts is invalid
+        runbook_url: {{ $.Values.defaults.runbookUrl }}#Alert-Rules-Selectors-Validity
+        description: >-
+          The DeploymentSelector used for deployment level alerts did not return any data.
+          Please check the DeploymentSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your deployments so that you
+          will be alerted for deployment issues. The current selector is
+          `{deploymentSelector="{{ $deploymentSelector }}"}`.
+      expr: |-
+        (
+          count(
+            kube_deployment_labels{
+              {{ $deploymentSelector }}
+            }
+          ) or on() vector(0)
+        ) == 0
       for: {{ .for }}
       labels:
         severity: {{ .severity }}
@@ -317,6 +372,33 @@ spec:
         {{- end }}
     {{- end }}
 
+    {{ with .StatefulsetSelectorValidity -}}
+    - alert: StatefulsetSelectorValidity
+      annotations:
+        summary: StatefulsetSelector for prometheus-alerts is invalid
+        runbook_url: {{ $.Values.defaults.runbookUrl }}#Alert-Rules-Selectors-Validity
+        description: >-
+          The StatefulsetSelector used for statefulset level alerts did not return any data.
+          Please check the StatefulsetSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your statefulsets so that you
+          will be alerted for statefulset issues. The current selector is
+          `{statefulsetSelector="{{ $statefulsetSelector }}"}`.
+      expr: |-
+        (
+          count(
+            kube_statefulset_created{
+              {{ $statefulsetSelector }}
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: {{ .for }}
+      labels:
+        severity: {{ .severity }}
+        {{- if $.Values.defaults.additionalRuleLabels }}
+        {{ toYaml $.Values.defaults.additionalRuleLabels | nindent 8 }}
+        {{- end }}
+    {{- end }}
+
     {{- end }}
     {{- end }}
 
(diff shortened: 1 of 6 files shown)


charts/prometheus-alerts/templates/containers-prometheusrule.yaml
```diff
@@ -512,4 +512,166 @@ spec:
     {{- end }}
     {{- end }}
 
+  - name: {{ .Release.Name }}.{{ .Release.Namespace }}.alertRulesSelectorsValidity
```
diranged, line 515: Can't these be inside the existing conditions for each group of alerts? Ie, this with the daemonset alerts?

charts/prometheus-alerts/templates/containers-prometheusrule.yaml
```diff
@@ -200,6 +200,34 @@ spec:
...
+    - alert: PodSelectorValidity
+      annotations:
+        summary: PodSelector for prometheus-alerts is invalid
+        runbook_url: {{ $.Values.defaults.runbookUrl }}#Alert-Rules-Selectors-Validity
+        description: >-
+          The PodSelector used for pod level alerts did not return any data.
+          Please check the PodSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your pods so that you
+          will be alerted for pod issues. The current selector is
+          `{podSelector="{{ $podSelector }}", namespaceSelector="{{ $namespaceSelector }}"}`.
```
diranged, line 213: I feel like this string is wrong ... shouldn't it just be `{ {{ $podSelector }}, {{ $namespaceSelector }} }`?

charts/prometheus-alerts/templates/containers-prometheusrule.yaml
```diff
@@ -200,6 +200,34 @@ spec:
...
+      expr: |-
```
diranged, line 214: Can you put some rendered output into the PR - that way we can quickly test it and sanity check.

charts/prometheus-alerts/templates/containers-prometheusrule.yaml
LaikaN57: I was not trying to put the exact promql. I was trying to map the prometheus-alerts values name to the rendered output to give people a searchable string in the values file and the actual promql.

LaikaN57:
```diff
--- orig        2024-04-25 18:22:00
+++ new 2024-04-25 18:20:37
@@ -47,7 +47,7 @@
     nextdoor.com/chart: prometheus-rules
     nextdoor.com/source: https://github.com/Nextdoor/k8s-charts
   labels:
-    helm.sh/chart: prometheus-alerts-1.5.0
+    helm.sh/chart: prometheus-alerts-1.6.0
     app.kubernetes.io/version: "0.0.1"
     app.kubernetes.io/managed-by: Helm
     
@@ -184,6 +184,29 @@
       for: 15m
       labels:
         severity: warning
+
+    - alert: PodSelectorValidity
+      annotations:
+        summary: PodSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The PodSelector used for pod level alerts did not return any data.
+          Please check the PodSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your pods so that you
+          will be alerted for pod issues. The current selector is
+          `pod=~"prometheus-alerts-.*", namespace="default"`.
+      expr: |-
+        (
+          count(
+            kube_pod_info{
+              pod=~"prometheus-alerts-.*",
+              namespace="default"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
+      labels:
+        severity: warning
     - alert: KubeDeploymentGenerationMismatch
       annotations:
         summary: Deployment generation mismatch due to possible roll-back
@@ -200,6 +223,28 @@
       for: 15m
       labels:
         severity: warning
+
+    - alert: DeploymentSelectorValidity
+      annotations:
+        summary: DeploymentSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The DeploymentSelector used for deployment level alerts did not return any data.
+          Please check the DeploymentSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your deployments so that you
+          will be alerted for deployment issues. The current selector is
+          `job="kube-state-metrics", kube_deployment=~"prometheus-alerts-.*", namespace="default"`.
+      expr: |-
+        (
+          count(
+            kube_deployment_labels{
+              job="kube-state-metrics", kube_deployment=~"prometheus-alerts-.*", namespace="default"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
+      labels:
+        severity: warning
     - alert: KubeStatefulSetReplicasMismatch
       annotations:
         summary: StatefulSet has not matched the expected number of replicas.
@@ -265,6 +310,28 @@
       for: 15m
       labels:
         severity: warning
+
+    - alert: StatefulsetSelectorValidity
+      annotations:
+        summary: StatefulsetSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The StatefulsetSelector used for statefulset level alerts did not return any data.
+          Please check the StatefulsetSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your statefulsets so that you
+          will be alerted for statefulset issues. The current selector is
+          `job="kube-state-metrics", statefulset=~"prometheus-alerts-.*", namespace="default"`.
+      expr: |-
+        (
+          count(
+            kube_statefulset_created{
+              job="kube-state-metrics", statefulset=~"prometheus-alerts-.*", namespace="default"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
+      labels:
+        severity: warning
     - alert: KubeDaemonSetRolloutStuck
       annotations:
         summary: DaemonSet rollout is stuck.
@@ -327,6 +394,28 @@
       for: 15m
       labels:
         severity: warning
+
+    - alert: DaemonsetSelectorValidity
+      annotations:
+        summary: DaemonsetSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The DaemonsetSelector used for daemonset level alerts did not return any data.
+          Please check the DaemonsetSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your daemonsets so that you
+          will be alerted for daemonset issues. The current selector is
+          `job="kube-state-metrics", daemonset=~"prometheus-alerts-.*", namespace="default"`.
+      expr: |-
+        (
+          count(
+            kube_daemonset_labels{
+              job="kube-state-metrics", daemonset=~"prometheus-alerts-.*", namespace="default"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
+      labels:
+        severity: warning
     - alert: KubeJobCompletion
       annotations:
         summary: Job did not complete in time
@@ -354,6 +443,28 @@
       for: 15m
       labels:
         severity: warning
+
+    - alert: JobSelectorValidity
+      annotations:
+        summary: JobSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The JobSelector used for job level alerts did not return any data.
+          Please check the JobSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your jobs so that you
+          will be alerted for job issues. The current selector is
+          `job="kube-state-metrics", job_name=~"prometheus-alerts-.*", namespace="default"`.
+      expr: |-
+        (
+          count(
+            kube_job_info{
+              job="kube-state-metrics", job_name=~"prometheus-alerts-.*", namespace="default"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
+      labels:
+        severity: warning
     - alert: KubeHpaReplicasMismatch
       annotations:
         summary: HPA has not matched descired number of replicas.
@@ -395,5 +506,27 @@
           ==
         kube_horizontalpodautoscaler_spec_max_replicas{job="kube-state-metrics", horizontalpodautoscaler=~"prometheus-alerts-.*", namespace="default"}
       for: 15m
+      labels:
+        severity: warning
+
+    - alert: HpaSelectorValidity
+      annotations:
+        summary: HpaSelector for prometheus-alerts is invalid
+        runbook_url: https://github.com/Nextdoor/k8s-charts/blob/main/charts/prometheus-alerts/runbook.md#Alert-Rules-Selectors-Validity
+        description: >-
+          The HpaSelector used for hpa level alerts did not return any data.
+          Please check the HpaSelector applied in your prometheus-alerts chart
+          is correct to ensure you are properly selecting your hpas so that you
+          will be alerted for hpa issues. The current selector is
+          `job="kube-state-metrics", horizontalpodautoscaler=~"prometheus-alerts-.*", namespace="default"`.
+      expr: |-
+        (
+          count(
+            kube_horizontalpodautoscaler_info{
+              job="kube-state-metrics", horizontalpodautoscaler=~"prometheus-alerts-.*", namespace="default"
+            }
+          ) or on() vector(0)
+        ) == 0
+      for: 1h
       labels:
         severity: warning
```

diranged (approved):

diranged (commented):
I just realized - the alerts do not sum by the namespace... which means they won't have anywhere to route to. Eg:

image

I think you need to explicitly add `namespace: {{ .Release.Namespace }}` to the `labels` key in each alarm

diranged (changes requested):
(sorry gotta fix that before we can merge)

diranged (approved):

PR #149, opened 2022-07-15
[prometheus-alerts/nd-common] Scope down the prometheus-alerts

diranged:
The prometheus-alerts chart was too widely scoped. This would make it
difficult to run multiple Applications within a single Namespace that
each define their own prometheus-alerts charts.

This change significantly breaks backwards compatibility - hence the
major version upgrade. The new code narrowly scopes the alerts by
default, and provides overrides to let a developer widen their scope.

PR #298, opened 2024-04-26
fix(prometheus-alerts): Make resource selectors more generic

LaikaN57:
We have changed the default selectors to be more generic. This means that if you deploy multiple applications to the same namespace, and one of those applications uses this chart, then by default all applications will be monitored by these alerts. You can change this behavior by modifying the `.Values.defaults.*NameSelector` regex values.

This is a more advantageous model as (like with the individual alerts themselves) it now requires opt-out (or more specific scoping) if the default is alerting on some objects you do not want to alert on.

Historical context:
- https://github.com/Nextdoor/k8s-charts/pull/149
- https://github.com/Nextdoor/k8s-charts/pull/150

### DIFF
--- a/charts/prometheus-alerts/Chart.yaml
+++ b/charts/prometheus-alerts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: prometheus-alerts
 description: Helm Chart that provisions a series of common Prometheus Alerts
 type: application
-version: 1.6.0
+version: 1.7.0
 appVersion: 0.0.1
 maintainers:
   - name: diranged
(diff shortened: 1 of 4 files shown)


scohen-nd:
Are those alerts attached per namespace or per simple-app? If it is per simple-app, how to we avoid the same problem that we have with istio-alerts where they override each other?
That's a more specific issue I noticed.

LaikaN57:
> Are those alerts attached per namespace or per simple-app? If it is per simple-app, how to we avoid the same problem that we have with istio-alerts where they override each other?

@scohen-nd This chart has no direct interaction with `simple-app`. I believe it is intended to use when someone is specifically not using `simple-app` (or similar).

LaikaN57:
> Are those alerts attached per namespace or per simple-app?

Per namespace. I think this chart is meant to be used specifically when an application is not deployed using `simple-app` (or similar).

diranged (approved):

PR #302, opened 2024-05-15
feat(charts/nd-common): allow option to omit DD_ENTITY_ID

tsingla1998:
We are trying to reduce ingested tags in the monolith by using a dogstatsd client that utilizes Vector instead of the Datadog Agent. The reason for this is because the Datadog Agent automatically includes some tags such has `host_id` and `instance_id`. By using vector we are able to more granularly control the global tags that are added to each metric. While we have made significant progress in reducing indexed tags cost, we are still ingesting all these tags which is becoming more significant portion of our metrics cost.

An issue I ran into while testing routing metrics via Vector is that the dogstatsd library reads `DD_ENTITY_ID` and will set that as a tag automatically if the environment variable is set. Right now, `container_id` is being set instead of reading from `DD_ENTITY_ID` and setting the `dd.internal.entity_id` tag. If this tag gets added automatically, we lose a lot of the potential gains since we have another unique tag that will explode cardinality. 

This PR proposes an opt-out for `DD_ENTITY_ID` and I would like to test that by opting out in our monolith flavors. I am hoping to get "cleaner" metrics with this variable being unset.

Example [test metric](https://app.datadoghq.com/metric/explorer?fromUser=true&start=1715800595409&end=1715803231679&paused=true#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYZAYgUJ4YThr42gDGSsgg6gi6mZaBZnHKGniqyCJwxAhQOG4AdAh4GowY2k4I5k4AbnCZCG5O+YXAAFQ8AAQARljTwFBQTaI92pFoTXAiEnJO0DxNGBpOmfi7ABQAlCA8ALpUru54mKHhz6qvGDGl8XePEAaORoHKgeQYUEIHrKBowM5vDQQTKJNCiODdBQ5IE4NFQVHopz0JjKERuDxoO78CD2TBYTGKZRomoAnh8IHyNEIADCUmEMBQIleaB4QA).

### DIFF
--- a/charts/daemonset-app/Chart.yaml
+++ b/charts/daemonset-app/Chart.yaml
@@ -2,12 +2,12 @@ apiVersion: v2
 name: daemonset-app
 description: Default DaemonSet Helm Chart
 type: application
-version: 0.15.5
+version: 0.15.6
 appVersion: latest
 maintainers:
   - name: diranged
     email: matt@nextdoor.com
 dependencies:
   - name: nd-common
-    version: 0.3.1
+    version: 0.3.2
     repository: file://../nd-common
--- a/charts/daemonset-app/README.md
+++ b/charts/daemonset-app/README.md
@@ -2,7 +2,7 @@
 
 Default DaemonSet Helm Chart
 
-![Version: 0.15.5](https://img.shields.io/badge/Version-0.15.5-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: latest](https://img.shields.io/badge/AppVersion-latest-informational?style=flat-square)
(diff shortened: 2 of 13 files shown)


diranged (commented):
I don't think that this is a terribly common feature - and we actually intend on helping to solve this in a different way with OTEL.. for the short term, can you just manually add the env variables exactly as they're defined below into your own charts, rather than modifying this common component? That should unblock you and let you test.

tsingla1998:
> I don't think that this is a terribly common feature - and we actually intend on helping to solve this in a different way with OTEL.. for the short term, can you just manually add the env variables exactly as they're defined below into your own charts, rather than modifying this common component? That should unblock you and let you test.

Sounds good. Thanks for taking a look.

PR #307, opened 2024-05-28
chore(prometheus-alerts): allow opting out of the validity selector alerts

diranged:
There are some cases where we know the selector is fine (ie, `.*`) and yet there are not ALWAYS going to be resources that match that selector - an example would be any `Job` that is created as part of a release process... Jobs may come and go, and there may be periods of times when a `Job` does not exist, so the `kube_job_info{}` metric won't exist.

In these cases, we need an escape hatch to allow the critical alarms to exist without alerting oncall engineers just because a job is _missing_..


diranged:
cc @LaikaN57 for visibility

LaikaN57 (commented):
Approved

PR #324, opened 2024-08-05
chore(charts/istio-alerts): update runbook for 5xx increase

schahal:
Due to encoding issues, the direct link to graph is not in the alerts...

So adding the query here for easier lookup

### DIFF
--- a/charts/istio-alerts/runbook.md
+++ b/charts/istio-alerts/runbook.md
@@ -27,6 +27,11 @@ sum by (destination_service_name, reporter) (
 )
 ```
 
+Action Items:
+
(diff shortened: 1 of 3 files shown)


charts/istio-alerts/runbook.md
```diff
@@ -10,6 +10,23 @@ service. Many services have custom dashboards in DataDog as well which may help
 investigate this alert further, and most service also produce logs of requests
 which may provide more context into what errors are being returned and why.
 
+Can check trends/graph by:
```
fastier-nextdoor, line 13: Should the runbook also provide action steps, such as: `if this, do that ; if this, ping that person ...`?

fastier-nextdoor (approved):

charts/istio-alerts/runbook.md
schahal: yes! added in https://github.com/Nextdoor/k8s-charts/pull/324/commits/81512a45f8ac7346214459cec9f0462a542d182d

PR #325, opened 2024-08-06
feat: allow using templating with terminationGracePeriodSeconds

stlava:
Adds template parsing on `terminationGracePeriodSeconds` in order to allow referencing or computing the termination value.


charts/daemonset-app/templates/tests/test-connection.yaml
```diff
@@ -3,8 +3,6 @@ apiVersion: v1
 kind: Pod
 metadata:
   name: "{{ include "nd-common.fullname" . }}-test-connection"
-  labels:
```
stlava, line 6: I don't know how this ever worked https://github.com/helm/helm/issues/8058

PR #334, opened 2024-09-11
feat: support adding priority class to workloads

scohen-nd:
Support adding priority class to workloads


charts/rollout-app/values.yaml
```diff
@@ -734,6 +734,10 @@ monitor:
   # -- ServiceMonitor will use these tlsConfig settings to make the health check requests
   tlsConfig: null
 
+# Set a different priority class to the pods, by default the default priority class is given to pods.
```
diranged, line 737: Change this to `# -- ...` and then helm-docs will turn this into a documentation comment in the README

diranged (approved):

PR #338, opened 2024-09-26
feat: enable maintenance mode (or custom HTTP fault injections)

schahal:
Similar to `retries`, this PR allows users to inject custom `faults` too to their `VirtualService`: https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPFaultInjection

We also support an opinionated fault injection called `maintenanceMode`: this will create a 100% abort fault injection to abstract away the implementation from end users (it also [implicitly deactivates the HPA](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#implicit-maintenance-mode-deactivation) while in maintenance mode)

The templates are smart enough to only render if either one of `maintenanceMode.enabled` or `faults` are set.

### DIFF
--- a/charts/simple-app/templates/deployment.yaml
+++ b/charts/simple-app/templates/deployment.yaml
@@ -7,6 +7,7 @@ make sure that each Deployment is scaled up separately from the next one.
 */}}
 {{- $deploymentZones := default (list "default") .Values.deploymentZones }}
 
+{{- $deactivateHpaForMaintenance := false }}
 
 {{- /*
 ...
@@ -27,6 +28,7 @@ Verify that some required inputs are supplied
 
 {{- if .Values.virtualService.enabled }}
 {{- $readinessProbe := required "readinessProbe is required" .Values.readinessProbe }}
+{{- $deactivateHpaForMaintenance = (and .Values.virtualService.maintenanceMode.enabled .Values.autoscaling.enabled) }}
 {{- end }}
 
 {{- /*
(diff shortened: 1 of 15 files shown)


charts/simple-app/templates/deployment.yaml
```diff
@@ -79,6 +81,9 @@ spec:
   {{- end }}
   {{- if and $.Values.replicaCount (not $.Values.autoscaling.enabled) }}
   replicas: {{ $.Values.replicaCount }}
+  {{- else if $deactivateHpaForMaintenance }}
+  {{- /* https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#implicit-maintenance-mode-deactivation */}}
+  replicas: 0
```
diranged, line 86: Wait ... I see the docs, but wouldn't this automatically scale the `Deployment` down? Have  you tested this behavior?

schahal: Looking at metrics of the deployment when we ran tests, it _**did**_ scale the `Deployment`  down:

Screenshot 2024-09-25 at 7 30 55 PM

As there are no documented ways to "pause" it (besides finagling with `behavior`), we _can_ try to use this approach mentioned in this comment: https://github.com/kubernetes/kubernetes/issues/28603#issuecomment-990507193

Its implementation is at https://github.com/Nextdoor/k8s-charts/pull/338/commits/d07c6c8902288ba5a67eb3abc9492346131f74fe

schahal: > besides finagling with behavior

Tested that we do _**not**_ indeed have to `merge` the behavior keys that come in - squashing in maintenance mode is good enough: https://github.com/Nextdoor/k8s-charts/pull/338/commits/0302e28110f68048d171a5d6ac668984f033b539

diranged (approved):

PR #340, opened 2024-09-27
(chore): Allow passing kafka and zookeeper resource limits via values file

philkjacobs-nextdoor:
Updates `charts/local-kafka` to allow passing in the resources object for kafka and zookeeper.


charts/local-kafka/values.yaml
```diff
@@ -24,6 +24,25 @@ kafka:
...
+
+zookeeper:
+  # -- Optional resource requirements for the Zookeeper cluster
+  resources:
+    limits:
+      cpu:
+      memory:
+    requests:
+      cpu:
+      memory:
```
LaikaN57, line 44: set to empty map by default

philkjacobs-nextdoor: Yes, that's intended. Or are you saying I shouldn't update this file to have the default-null keys at all?

PR #342, opened 2024-10-25
feat: analogous AuthorizationPolicies to NetworkPolicies for (Istio) multi-cluster access

schahal:
## What this change does

In essence:

- If service has Istio enabled, adds `AuthorizationPolicy` that is _**analogous**_ to the existing `NetworkPolicy` that we already create (**this should be a no-op**).
- If service also enables ~`.Values.network.multiCluster.allowFromRemote`~  `.Values.network.allowAll`, we open up the NetworkPolicy to allow the AuthorizationPolicies above to take over its enforcement wholly.

## Why

To enable services in a multi-primary (Istio mesh), mult-cluster setup to be accessed from services in a different cluster (because NetworkPolicies can't enforce Ingress from **outside** the Kubernetes cluster without getting to CIDR ranges - i.e., it only knows about cluster-local namespaces. So, we allow all and instead restrict with Istio's AuthorizationPolicy)

### DIFF
--- a/charts/nd-common/templates/_authorizationpolicy.tpl
+++ b/charts/nd-common/templates/_authorizationpolicy.tpl
@@ -1,0 +1,81 @@
+{{- /*
+
+This function creates the following two AuthorizationPolicy objects:
+
+  1. To allow same-namespace access (this can probably be migrated to a
+     Kyverno ClusterPolicy that applies this on all namespaces, but for
+     now adding here for smooth transition for "allow" AuthorizationPolicies
+     to be created too)
+
+  2. To allowNamespaces to have ingress access to the service (a drop-in
+     replacement of the NetworkPolicy we make defunct when a service is to
+     be accessed from a multi-cluster setup
+
+These objects are generally pretty simple, but we re-use them in a few places
+and it's nice to have one common way to make them.
+
+AuthorizationPolicies can be used in lieu of NetworkPolicies in a multi-
+cluster setup
+
+Via https://istio.io/latest/docs/concepts/security/#allow-nothing-deny-all-and-allow-all-policy:
+
+> Note the “deny by default” behavior applies only if the workload has at least one authorization
+policy with the ALLOW action.
+
+- */}}
+{{- define "nd-common.authorizationPolicy" }}
+{{- if .Values.istio.enabled }}
+{{- /*
+
+Create a default AuthorizationPolicy that allows local namespace ingress
+
+See note above: after a while, wWe can probably have this as part of a
+Kyverno ClusterPolicy that's added to all namespaces.
+
+- */}}
+---
+apiVersion: security.istio.io/v1beta1
+kind: AuthorizationPolicy
+metadata:
+  name: allow-local-namespace-ingress
+spec:
+  selector:
+    matchLabels:
+      {{- include "nd-common.selectorLabels" . | nindent 6 }}
+  action: ALLOW
+  rules:
+  - from:
+    - source:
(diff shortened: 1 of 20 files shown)


charts/nd-common/templates/_authorizationpolicy.tpl
```diff
@@ -0,0 +1,81 @@
...
+
+See note above: after a while, wWe can probably have this as part of a
+Kyverno ClusterPolicy that's added to all namespaces.
+
+- */}}
+---
+apiVersion: security.istio.io/v1beta1
+kind: AuthorizationPolicy
+metadata:
+  name: allow-local-namespace-ingress
```
diranged, line 40: This should not happen  here… let’s talk on Monday about this and why…

charts/nd-common/templates/_authorizationpolicy.tpl
```diff
@@ -0,0 +1,81 @@
...
+spec:
+  selector:
+    matchLabels:
+      {{- include "nd-common.selectorLabels" . | nindent 6 }}
+  action: ALLOW
+  rules:
+  - from:
+    - source:
+        namespaces:
+        {{- range .Values.network.allowedNamespaces }}
```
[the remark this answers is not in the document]
schahal: This works, confirmed it'll quote when passing in an asterisk (`*`):

```diff
@@ -433,8 +433,9 @@
   - from:
     - source:
         namespaces:
-        - "foo"
-        - "bar"
+        - foo
+        - bar
+        - '*'
     to:
     - operation:
         ports:
```
.. which is the desired behavior. Updated in last commit.

charts/nd-common/templates/_authorizationpolicy.tpl
```diff
@@ -0,0 +1,52 @@
...
+spec:
+  selector:
+    matchLabels:
+      {{- include "nd-common.selectorLabels" . | nindent 6 }}
+  action: ALLOW
+  rules:
+  - from:
+    - source:
+        namespaces: [{{ .Release.Namespace }}]
+{{- if .Values.ports }}
```
diranged, line 35: You don't have to fix this - but my recommendation to make these things easier to read is to make sure to indent the `{{ if ..}}.. {{ end }}` statements so that you don't end up with this:

```
{{- end }}
{{- end }}
{{- end }}
{{- end }}
{{- end }}
```

Also, this could be turned into a single If statement.

charts/simple-app/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-app
 description: Default Microservice Helm Chart
 type: application
-version: 1.11.1
+version: 1.11.2
```
diranged, line 5: Seems like we should bump to 1.12.0

charts/stateful-app/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: stateful-app
 description: Default StatefulSet Helm Chart
 type: application
-version: 1.3.1
+version: 1.3.2
```
diranged, line 5: 1.4.0 .. since its a new feature..

diranged (commented):
Just a few notes on bumping the versions up.. thats it!

charts/nd-common/templates/_authorizationpolicy.tpl
schahal: Yea since we're here, let's fix it (I took the multi-pattern from the networkpolicy, but single-line is more readable).

Updated.

diranged (approved):

PR #344, opened 2024-10-30
chore: auto-add istio ingress gateway namespace to AuthorizationPolicy

schahal:
Ref: https://github.com/Nextdoor/k8s-charts/pull/342

## Motivation

We have an `AuthorizationPolicy` where users can set allowedNamespaces. However, it's easy to miss the namespace of the ingress gateway that users may want their VirtualService tied to.

## What This Does

This parses the namespace of the istio gateways that users provide and ~creates a new list by appending it to the allowedNamespaces-provided list~ adds a separate rule for it

This is a safe assumption due to our checks of its format: https://github.com/Nextdoor/k8s-charts/blob/ca85ca6dedd25eb979e4f3f8240bd53cc424fa45/charts/simple-app/templates/istio/virtualservice.yaml#L19

## Test

Updated set `network.allowAll` and `virtualService.enabled` in values.local.yaml to `true...

Resulting relevant diff of `make template`:

### First Commit

```diff
--- /tmp/simple-app-orig.yaml	2024-10-29 19:55:03
+++ /tmp/simple-app-new.yaml	2024-10-29 19:57:33
...
@@ -424,7 +424,7 @@
 metadata:
   name: simple-app
   labels:
-    helm.sh/chart: simple-app-1.12.1
+    helm.sh/chart: simple-app-1.12.2
     app.kubernetes.io/version: "latest"
     app.kubernetes.io/managed-by: Helm
     tags.datadoghq.com/service: "simple-app"
@@ -466,6 +466,7 @@
         namespaces:
         - foo
         - bar
+        - istio-gateways
     to:
     - operation:
         ports:
...
```

### Updated Commit

```diff
...
@@ -471,6 +471,15 @@
         ports:
         - "80"
         - "8443"
+  - from:
+    - source:
+        namespaces:
+        - "istio-gateways"
+    to:
+    - operation:
+        ports:
+        - "80"
+        - "8443"
 ---
...
```

### DIFF
--- a/charts/nd-common/templates/_authorizationpolicy.tpl
+++ b/charts/nd-common/templates/_authorizationpolicy.tpl
@@ -16,6 +16,22 @@ Via https://istio.io/latest/docs/concepts/security/#allow-nothing-deny-all-and-a
 policy with the ALLOW action.
 
 - */}}
+
+{{- define "nd-common.allAllowedNamespaces" -}}
+  {{- $res := .Values.network.allowedNamespaces -}}
+  {{- /*
+    Start off with allowedNamespaces, then append istio
+    ingress gateway namespaces
+  */}}
+  {{- if .Values.virtualService.enabled -}}
+    {{- range .Values.virtualService.gateways -}}
+      {{- $gwParts := splitList "/" .  -}}
+      {{- $res = append $res (first $gwParts) -}}
+    {{- end -}}
+  {{- end -}}
+{{- $res | uniq | toYaml -}}
+{{- end -}}
+
 {{- define "nd-common.authorizationPolicy" }}
 {{- if and .Values.istio.enabled (.Capabilities.APIVersions.Has "security.istio.io/v1beta1") }}
 ---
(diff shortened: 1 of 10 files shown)


charts/nd-common/templates/_authorizationpolicy.tpl
```diff
@@ -32,11 +48,14 @@ spec:
   - from:
     - source:
         namespaces: [{{ .Release.Namespace }}]
-  {{- if and .Values.ports (gt (len .Values.ports) 0) (gt (len .Values.network.allowedNamespaces) 0) }}
+  {{- if and
```
diranged, line 51: This seems a little complicated… how about just a second rule that handles the istio gateway rules separately from the other ones?

schahal: Originally, I believed the de-dupe of the ports would've gotten a bit wordy, but comparing the two, I also thing the separate rule (latest commit) is a bit simpler aesthetically (?)

diranged (approved):

PR #347, opened 2024-11-05
feat: Adding new datadog monitors chart

scohen-nd:
Adding new datadog monitors chart to allow users to use it as a subchart and manage datadog alerts via the datadog CRD's.


charts/datadog-monitors/values.yaml
```diff
@@ -0,0 +1,83 @@
...
+    # - error-tracking: `error-tracking alert`
+    # - database-monitoring: `database-monitoring alert`
+    # - network-performance: `network-performance alert`
+    # - service-discovery: `service-discovery alert`
+    type: "query alert"
+    # -- (`string[]`) Optional: Additional monitor tags(will be added on top of the default tags:service, team, namespace, cluster, app_env, app_group)
+    tags:
+      - 'tagname:tagvalue'
+    # -- (`map[string]interface{}`) Optional: monitor options
+    options:
```
LaikaN57, line 49: I would ignore all the sub-keys here and just point to the upstream documentation instead as we are not likely to keep this list up to date.

charts/datadog-monitors/values.yaml
```diff
@@ -0,0 +1,83 @@
...
+serviceName:
+# -- (`string`) Optional shared pagerduty team name for monitors, will turn to a tag for alerts - if not provided, the tag will not be added
+team:
+
+# -- (`map[string]interface{}`) List of monitors
+monitors:
+  # -- (`map[string]interface{}`) Required: monitor resource name, Required unique monitor resource name(needed to allow value overrides and used a datadog monitor resource name)
+  resourceName:
+    # -- (`boolean`) Optional: whether to exclude the monitor, defaults to False
+    disabled: true
```
LaikaN57, line 11: Typically in k8s we use the `enabled` flag so we don't end up with double negatives.

charts/datadog-monitors/values.yaml
```diff
@@ -0,0 +1,83 @@
+# -- (`string`) Optional shared pagerduty service name for monitors, will turn to a tag for alerts - if not provided, the .Release.name will be used by default
+serviceName:
+# -- (`string`) Optional shared pagerduty team name for monitors, will turn to a tag for alerts - if not provided, the tag will not be added
+team:
+
+# -- (`map[string]interface{}`) List of monitors
+monitors:
+  # -- (`map[string]interface{}`) Required: monitor resource name, Required unique monitor resource name(needed to allow value overrides and used a datadog monitor resource name)
+  resourceName:
```
LaikaN57, line 9: I would not create a real entry here and instead just have a large comment block. (Although I see you have disabled the entry below so I dont have strong opinions on this.)

LaikaN57 (approved):
Left some non-blocking comments.

charts/datadog-monitors/values.yaml
scohen-nd: That was my original intention, but couldn't figure out how to properly format it so that helm-docs recognize the fields....

charts/datadog-monitors/values.yaml
scohen-nd: It's an option I thought about for the whole thing, in theory the names and the structure is exactly the same as the CRD - I could in theory point out to the CRD(though some of the info was not properly documented and I had to extract it from the code).

charts/datadog-monitors/values.yaml
scohen-nd: I reversed that logic to use enabled instead of disabled.

PR #348, opened 2024-11-08
chore: Fixing datadog-monitor chart docs

scohen-nd:
Fixing/aligning datadog-monitor chart docs.


LaikaN57 (approved):
Looks like a few truthy warning but other that, gtg

PR #354, opened 2025-01-10
chore(nd-common): add support for goldilocks workload labels

scohen-nd:
Adding support for goldilocks labels to enable and select the VPA update mode per workload.

The following structure is supported:
```
goldilocks:
   enabled: false/true  # defaults to true if goldilocks values are defined
    updateMode: off/initial/auto  # defaults to off
```


charts/nd-common/templates/_goldilocks.tpl
```diff
@@ -0,0 +1,17 @@
+{{/*
+
+The "goldilocksLabels" function creates common Goldilocks labels that can be applied
+to workloads(like: deployment, statefulset, etc...). These labels help configure the Goldilocks VPAs. This
+function automatically checks if `.Values.datadog.enabled` is True, so you do
```
diranged, line 5: Typo - not datadog

charts/nd-common/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: nd-common
 description: A helper chart used by most of our other charts
 type: library
-version: 0.3.6
+version: 0.3.7
```
diranged, line 5: This is a new feature - should move to `0.4.0`

diranged (changes requested):

charts/daemonset-app/Chart.yaml
```diff
@@ -2,12 +2,12 @@ apiVersion: v2
 name: daemonset-app
 description: Default DaemonSet Helm Chart
 type: application
-version: 0.16.2
+version: 0.16.3
```
diranged, line 5: So we've updated the way `nd-common` works - but we haven't updated the `values.yaml` (which in-turn updates the README) to help our users actually know about the `.Values.goldilocks.*` settings. While it's a little tedious, I think we should update the `values.yaml` files in each of our apps here so that the user is aware of these settings. I also think that means bumping the version of these charts by a major rev as well (0.16 -> 0.17) - thoughts?

scohen-nd: So, replacing the default logic with default values for documentation purposes?
Make sense to do that, I was wondering why it was not done so far.
We did not add a breaking change nor a major change to the deployment, statefulset or daemonset, that's why I did not think it requires a major version upgrade.

scohen-nd: Done.

diranged (approved):
looks good to me.. thanks!

PR #355, opened 2025-01-21
feat(local-kafka): switch to kraft, start faster, allow annotations

diranged:
## Goal

The goal for this PR is to improve the usability of the `local-kafka` chart by reducing its startup time, introducing podAnnotations and removing old legacy settings:

### Change: Removed legacy startup/readiness probe delays

We no longer see the Strimzi Operator take minutes to start up on M1 laptops, so I am removing the arbitrary startup/readiness probe delays.

### Change: Replace Zookeeper with KRaft

Strimzi is moving to removing Zookeeper entirely in the 0.46.0 release anyways ... this change gets our local testing setup that way ahead of the change. This also reduces the number of pods that have to run locally, reducing the total startup time of a full testing environment.

### Change: Introduce `podAnnotations` and `annotations` settings

If you set `.Values.podAnnotations: { ... }` now we will add these annotations to the Kafka and Entity Operator pods. This makes it easier to tell Helm to start these pods up before the rest of the tested resources get created.

### DIFF
--- a/charts/local-kafka/values.yaml
+++ b/charts/local-kafka/values.yaml
@@ -9,7 +9,7 @@ clusterName: default
 # put into is used. It is recommended to keep this empty.
 namespaceOverride:
 
-# -- Optionall annotations added to all of the resources managed by this template.
+# -- Optional annotations added to all of the resources managed by this template.
 annotations: {}
 
 # -- Optional annotations added to all of the Strimzi-managed Pods.
--- a/charts/local-kafka/values.yaml
+++ b/charts/local-kafka/values.yaml
@@ -9,6 +9,9 @@ clusterName: default
 # put into is used. It is recommended to keep this empty.
 namespaceOverride:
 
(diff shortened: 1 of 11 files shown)


charts/local-kafka/values.yaml
```diff
@@ -9,6 +9,12 @@ clusterName: default
 # put into is used. It is recommended to keep this empty.
 namespaceOverride:
 
+# -- Optionall annotations added to all of the resources managed by this template.
```
NaturalFractals, line 12: super nit: extra `l` on Optional.

NaturalFractals (approved):
Awesome! lgtm

LaikaN57 (approved):

charts/local-kafka/values.yaml
diranged: :+1: thanks!

diranged: Typo fixed

PR #359, opened 2025-01-31
fix(istio-alerts): Fix istio alerts conflict issue

scohen-nd:
Our istio-alerts implementation is suffering from an issue with helm that does not pass the alias as the chart name to the sub sub chart, this leads to all the istio-alerts to have the .Chart.Name of "istio-alerts", while their parent chart gets the alias.
Example:
```
Parent chart:
  simple-app: alias child1
  statefulset-app: alias child2
```
both `child1` and `child2` passed on properly to `simple-app` and `statefulset-app`, but not to their subcharts - leaving their subcharts with the same name(hence: `istio-alerts`) - when that it used in a template it generates templates with the same name, when deployed to the same namespace they override one each other.
By default we had this issue and someone changed the default service name for the alerts to be `.*`, instead of empty as the existing logic implied - making it a namespace based rule and not allowing per app customization.

This change is attempting to use .Template.Name to extract the alias of the parent if the .Chart.Name is `istio-alerts` and use it in both the PrometheusRule name and in the destination_service to select the specific service (while allowing the users to set a more specific destinationService attribute in case it is different from the alias name).

### DIFF
--- a/charts/istio-alerts/templates/_helpers.tpl
+++ b/charts/istio-alerts/templates/_helpers.tpl
@@ -1,3 +1,14 @@
+{{- define "istio-alerts.alias" -}}
+  {{- if eq .Chart.Name "istio-alerts" -}}
+    {{- $path := splitList "/" .Template.Name -}}
+    {{- if gt (len $path) 3 -}}
+    {{- (index $path 2) | lower -}}
+    {{- end -}}
+  {{- else -}}
+    {{- .Chart.Name | lower -}}
+  {{- end -}}
+{{- end -}}
+
 {{- define "istio-alerts.namespaceSelectorIstioForMetrics" -}}
 destination_service_namespace="{{ .Release.Namespace }}"
 {{- end }}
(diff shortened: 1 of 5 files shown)


charts/istio-alerts/values.yaml
```diff
@@ -23,7 +23,7 @@ serviceRules:
   # -- Narrow down the alerts to a particular Destination Service if there
   # are multiple services that require different thresholds within the same
   # namespace.
-  destinationServiceName: .*
+  destinationServiceName: ""
```
schahal, line 26: If I understand this correctly, combined with the if/else change in _helpers.tpl, this will add the newly created istio-alerts alias into the destination_service_name selector for the query...

Unless, I'm misunderstanding the change, this is _not_ what we'd want, right? (because `destination` for any of these workloads can be anything).... so my understanding is keep the destination selector changes as they were (?)

(again, want to underscore, I may be missing some context here so please correct me).

scohen-nd: So, there is no way for the istio-alerts to knwo the exact name of the service, but by default it uses the chart-name which is the alias as the service name. At least based on the few services that I looked at.
Using that as a fallback is better than using ".*" which means the same alert will fire multiple times due to the multiple prometheus rules.
An alternative - we could require to fail any time "destinationService" is not set, but that means failing all our deployments until people change it - not sure there will be a go for such breaking change.

charts/istio-alerts/templates/_helpers.tpl
```diff
@@ -1,10 +1,23 @@
+{{- define "istio-alerts.alias" -}}
+  {{- if eq .Chart.Name "istio-alerts" -}}
+    {{- $path := splitList "/" .Template.Name -}}
+    {{- if gt (len $path) 3 -}}
+    {{- (index $path 2) | lower -}}
+    {{- end -}}
+  {{- else -}}
+    {{- .Chart.Name | lower -}}
+  {{- end -}}
+{{- end -}}
```
LaikaN57, line 10: Does this work for cases using `simple-app` and cases not using `simple-app`?

scohen-nd: Yes, I tested it in couple of scenarios.

PR #376, opened 2025-06-12
fix(nd-common): Use fullname for consistencey

LaikaN57:
So we have some weird inconsistencies when you try to set both `nameOverride` and `fullNameOverride`. It seems like we should just standardize on `fullName`. But we may need to do a bunch of test rendering to ensure we don't break any downstreams.

https://github.com/search?q=org%3ANextdoor+nd-common.name++NOT+is%3Aarchived+NOT+repo%3ANextdoor%2Fk8s-charts&type=code


diranged (changes requested):
This is going to be a breaking change - selector labels are consider immutable.

PR #381, opened 2025-06-13
chore(ci): Upgrade chart-releaser to v1.7.0

LaikaN57:
Args changed a bit: https://github.com/helm/chart-releaser-action/tree/v1.7.0?tab=readme-ov-file#inputs

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,10 +27,9 @@ jobs:
           version: v3.4.0
 
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.6.0
+        uses: helm/chart-releaser-action@v1.7.0
         with:
           charts_dir: charts
-          charts_repo_url: https://k8s-charts.nextdoor.com/
+          skip_existing: true
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -27,10 +27,9 @@ jobs:
           version: v3.4.0
 
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.6.0
+        uses: helm/chart-releaser-action@v1.7.0
         with:
           charts_dir: charts
-          charts_repo_url: https://k8s-charts.nextdoor.com/
```
LaikaN57, line 33: > pages_branch: Name of the branch to be used to push the index and artifacts. (default to: gh-pages but it is not set in the action it is a default value for the chart-releaser binary)

.github/workflows/release.yaml
```diff
@@ -27,10 +27,9 @@ jobs:
...
+          skip_existing: true
```
LaikaN57, line 33: > skip_existing: Skip package upload if release/tag already exists

diranged (approved):

PR #384, opened 2025-07-03
feat: integrated capabilities for multiple zones into charts/rollout-app

emoral435:
> [!IMPORTANT]
> For reviewers, I suggest going over the `rollout.yaml` file first to get a better understanding of how zones are handled, what is checked, etc. so that when reviewing the diffs to the other templates, things make a bit more sense 🙂
## Context | Description | Why
Currently, if a user wanted to add multiple zones / add a Rollout operator for different clusters, they would have to have multiple instances of a Rollout CRD definition within a file / have multiple files for each rollout on top of having to manually change and keep track of certain fields within the `spec` in order to customize and make each rollout service have different selectors.

This PR changes that, integrating the field `rolloutZones` to automatically make Rollout operators for each specific zone, creating duplicate rollout services for each one with other configurable options. This is handy to have, for instance, if one of your configured zones handles _substantially_ more traffic at times, and thus, can make it such that splitting up the rollout between zones can help save some costs.

## Testing
* running helm template showed no errors!
* I `helm install`'ed two versions of this rollout-app: one with an empty `rolloutZones: []`, and it worked, and I also installed a version of the app where the rolloutZones was `rolloutZones: [us-west-2a]`, added an appropriate label to a node selector, and it worked! So, rollout zones may need more configuration to get the node selection to work, but otherwise, this should _not_ be a breaking change
* I also tested out the `rolloutZonesTransition` variable, and it worked! Below is an example of how it works
* I also tested this out within the `dev1-test` cluster with a dummy repo on a branch. To see the branch and testing, contact me so that I can send it via slack! (since this is a public repo, I don't want to share it here, haha)

## Videos
Contact me so that I can send the video of it working as well through Slack!


charts/rollout-app/templates/rollout.yaml
```diff
@@ -1,182 +1,262 @@
...
+  name: {{ $fullName }}
   labels:
-    {{- include "rollout-app.labels" . | nindent 4 }}
+    {{- include "rollout-app.labels" $ | nindent 4 }}
+    {{- with $rolloutZoneLabel }}
+    {{ . }}
+    {{- end }}
 spec:
-  {{- with .Values.minReadySeconds }}
+  {{- with $.Values.minReadySeconds }}
```
teastburn, line 72: I had to add the `$` before all of the `.Values` references to get this to work, but i'm not actually sure if that's the best way to do this. @schahal would know

charts/rollout-app/templates/services.yaml
```diff
@@ -60,13 +59,11 @@ spec:
 This optional "Canary" service is created only when the .Values.strategy is set
 to "canary". Rollouts will then use this service to route canary traffic to, in
 order to test releases.
-
 When using Istio VirtualServices (if .Values.virtualService.enabled=true), the
```
teastburn, line 62: Is there any case where if someone wanted to have 1 rollout per AZ, that they'd also want 1 virtual service per AZ? I think the "umbrella" service concept would be needed as well so that the region's service as a whole can have a single ingress URL for clients to access it from other services. Otherwise the client has to know which AZ they're in and change the host in the URL for that AZ, which we don't do currently.

charts/rollout-app/templates/rollout.yaml
schahal: Correct, since we're in a `range` and `.Values.*` is at the top-level, to reset back at the top-level, makes sense to us `$.Values.*`(otherwise helm/go template rendering wouldn't find it).

Which I see this line is already doing, right?

emoral435:
Awesome: I removed the unused variable within the `hpa.yaml` file, added the if statement within the top of the `pdb` file, and the VirtualService logic now ensures that there are no breaking changes with current rollout users, but for future adopters of the multiple zones features, we ensure that we don't spin up multiple VirtualServices 🚀

PR #387, opened 2025-07-15
feat(simple-app): added rollouts (argo-rollouts) to simple-app

emoral435:
## Context
Currently, the `simple-app` and `rollout-app` have _very_ similar capabilities and even code structure at the moment, so with this PR, this merges the two to be within just the `simple-app`. By default, the services using `simple-app` will not be using a rollout deployment schema, however, if the user has the variable `rollout.enabled` set to `true`, then the user will be able to configure their deployment schema to be a rollout deployment (handled by [Argo Rollouts](https://argoproj.github.io/rollouts/)).

> [!IMPORTANT]
> For now, we will be leaving the `rollout-app` within the repository. This PR will mark the `rollout-app` as deprecated. However, due to the charts being similar, for future reference, the chart will be deleted within a following release of `k8s-charts.git`. This is due to the previously mentioned similarity between the two, and also to force developers to _actively be aware_ that their services may _benefit greatly_ in using rollout deployments! It will also be very confusing for developers to have `simple-app` have rollouts, whilst there is also a `rollout-app`
>
> Tier 0 and usually even tier 1 services should use rollouts deployments. All other services are at their own disgression to use rollouts or not, depending on the complexity of their services, the severity of safety revolving around the users the service is for, etc.

### DIFF
--- a/charts/simple-app/Chart.yaml
+++ b/charts/simple-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: simple-app
 description: Default Microservice Helm Chart
 type: application
-version: 1.13.3
+version: 1.14.0
 appVersion: latest
 maintainers:
   - name: diranged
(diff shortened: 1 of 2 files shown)


emoral435:
Closing–did not get to the merge of these two apps in time, but this is still an issue that will be continued potentially in the future.

PR #393, opened 2025-07-18
feat(rollout-app): added Datadog tags for stable and canary rollout ReplicaSets

emoral435:
## Context
Currently, if you have a DataDog agent running on your k8s cluster and your Argo Rollouts controller is [configured correctly](https://docs.datadoghq.com/integrations/argo-rollouts/) to allow the agent to scrape the `/metrics` endpoint for rollout metrics, you can get 90% of the granularity that you might need for the creation of a dashboard.

However, [DataDog allows](https://docs.datadoghq.com/containers/kubernetes/tag/?tab=datadogoperator) annotations to be added onto specific _ReplicaSets_ themselves for the creation of tags. These tags can have `KEY:VALUE` pairings for increased granularity into observability dashboards. This change in specific allows you to toggle between stable and canary ReplicaSets on a dashboard.

## Testing
One of my applications is currently running on these changes, and has the correct annotations applied to them. On top of that, I also see that the DataDog agent is able to successfully send the tag `KEY:VALUE` pairings to my dashboard 🙌 !

### DIFF
--- a/charts/rollout-app/templates/rollout.yaml
+++ b/charts/rollout-app/templates/rollout.yaml
@@ -36,7 +36,6 @@ disable this.
 {{/* The default rolloutZoneLabel is left empty. We patch it as we loop through $rolloutZones */}}
 {{- $rolloutZoneLabel := "" }}
 
-
 {{/* Iterate through the rollout zones now */}}
 {{- range $rolloutZone := index $rolloutZones }}
 
@@ -167,14 +166,16 @@ spec:
         {{- end }}
       {{- end }}
 
-      {{- with $.Values.canary.canaryMetadata }}
+      {{- $canaryMetadata := include "rollout-app.canaryMetadata" $ | fromYaml }}
+      {{- if $canaryMetadata }}
       canaryMetadata:
-        {{- toYaml . | nindent 8 }}
+        {{- toYaml $canaryMetadata | nindent 8 }}
       {{- end }}
 
-      {{- with $.Values.canary.stableMetadata }}
+      {{- $stableMetadata := include "rollout-app.stableMetadata" $ | fromYaml }}
+      {{- if $stableMetadata }}
       stableMetadata:
-        {{- toYaml . | nindent 8 }}
+        {{- toYaml $stableMetadata | nindent 8 }}
       {{- end }}
 
       {{- with $.Values.canary.maxUnavailable }}
--- a/charts/rollout-app/Chart.yaml
+++ b/charts/rollout-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: rollout-app
 description: Argo Rollout-based Application Helm Chart
 type: application
-version: 1.5.3
+version: 1.5.4
 appVersion: latest
 maintainers:
   - name: diranged
--- a/charts/rollout-app/README.md
+++ b/charts/rollout-app/README.md
@@ -2,7 +2,7 @@
 
 Argo Rollout-based Application Helm Chart
 
-![Version: 1.5.3](https://img.shields.io/badge/Version-1.5.3-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: latest](https://img.shields.io/badge/AppVersion-latest-informational?style=flat-square)
+![Version: 1.5.4](https://img.shields.io/badge/Version-1.5.4-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: latest](https://img.shields.io/badge/AppVersion-latest-informational?style=flat-square)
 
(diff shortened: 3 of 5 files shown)


charts/rollout-app/templates/rollout.yaml
```diff
@@ -36,6 +36,17 @@ disable this.
...
 
+{{/*
+If the user configured DataDog to be enabled, then we want to add annotations onto the ReplicaSet pods
+to help differentiate within DataDog dashboards between canary and stable ReplicaSets. This variables
+helps with ensuring that we only inject the annotations once, and don't override any additional annotations
+that the user may want to have added. For more information on adding tags using annotations, see:
+https://docs.datadoghq.com/containers/kubernetes/tag/?tab=datadogoperator
+*/}}
+{{- $rsAnnotationsAdded := false}}
+{{- $ddRSTag := "argo_rollouts_replicatset_type" }}
```
schahal, line 47: spelling :)

charts/rollout-app/templates/rollout.yaml
```diff
@@ -167,15 +178,54 @@ spec:
...
 
-      {{- with $.Values.canary.canaryMetadata }}
+      {{- /*
+      Range over canary and stable metadata fields that the user specified.
+      This logic makes sure that no fields get overridden if they have Datadog enabled,
+      while still injecting the necessary Datadog tags.
+      */ -}}
+      {{- $rsAnnotationsAdded = false }}
       canaryMetadata:
-        {{- toYaml . | nindent 8 }}
```
schahal, line 172: I think removing this and handling the YAML structure on our own (a) can be hard to to read, (b) bubble up some bugs.

It may be safer and more readable to use helm's `dict()` and then `merge()` function.

1. We have a `_helpers.tpl` file in this directory
2. There, define new variable(s) which is basically a `merge()` of the original annotations and what you have here already... e.g.,
        
        # define/set appropriate metadata (e.g., canary's or stable's) somewhere here...
        # [...]
        # Then, something like: 
        {{- $ddAnnotations := dict "ad.datadoghq.com/tags" (printf "%s: %s" "argo_rollouts_replicaset_type" "canary") -}}
        {{- $metadata = merge $metadata (dict "annotations" (merge ($metadata.annotations | default dict) $ddAnnotations)) -}}

End result is a variable that you can use in this file, that would basically be:

```
{{- with $.Values.canary.stableMetadata }}
stableMetadata:
  {{- include <whateverVarsYouCreated> | nindent  8}}
{{- end }}
```

schahal (changes requested):

charts/rollout-app/templates/rollout.yaml
emoral435: Oh dang! Nice catch, repli`CAT` 😹 guess this is a fruedian slip that I am missing my 15 cats back at home! Fixed 👍

charts/rollout-app/templates/rollout.yaml
```diff
@@ -167,14 +166,14 @@ spec:
         {{- end }}
       {{- end }}
 
-      {{- with $.Values.canary.canaryMetadata }}
+      {{- if or $.Values.canary.canaryMetadata $.Values.datadog.enabled }}
```
schahal, line 169: Can this somehow simply be some form of "if rollout-app.canaryMetadata"? <--- because that var handles the `or` logic in that definition.

schahal (changes requested):

charts/rollout-app/templates/rollout.yaml
emoral435: I tested this out right now, and it seems like most conventual ways of checking whether the map was empty (the way you recommended and also checking its length, storing it as a separate variable right above and then doing the comparison) still evaluated it as **true** when I would use Helm template locally, which is super weird, as the documentation regards empty maps as [falsy](https://helm.sh/docs/chart_template_guide/control_structures/#ifelse) 🧐

The current if statement flow does not render empty maps, so should we not keep it this way? I am not super opinionated here though!

emoral435:
Changed the README.md file and also added some more comments in `_helper.tpl` to explain the flow a bit better 🙂

charts/rollout-app/templates/rollout.yaml
emoral435: Thank god weekends exist: **I get what you mean now**. I was having a brain fart–since what I wrote in `_helper.tpl` returns toYaml (a string) it would basically always be evaluated as true. The newest changes now uses `fromYaml` to do the comparison you mentioned. Looks much cleaner now 👍

schahal (changes requested):
> Changed the README.md 

README.md is actually auto-generated from the README.gotmpl file (so the changes should go in the latter, otherwise it will be overwritten)

emoral435:
Nice catch! Just changed that up. The correct `README` file is now updated, and lets users know the new DD annotations are automatically applied just when they set the `Values.datadog.enabled` flag to true 🐘

schahal (approved):

PR #394, opened 2025-07-18
feat(charts/nd-common): add support to override common labels

saansh45:
### Problem

Helm users occasionally override `nd-common.labels` by directly appending additional labels like:

```yaml
labels:
  {{- include "nd-common.labels" . | nindent 4 }}
  tags.datadoghq.com/service: nextdoor-tw
  tags.datadoghq.com/version: release
```

However, when those keys (`tags.datadoghq.com/service`, etc.) already exist in the output of `nd-common.labels`, it results in a **duplicate mapping key YAML error**, such as:

```
YAMLException: duplicated mapping key (17:5)
```

This breaks `helm template` and `helm diff`, making the chart unusable when duplicate keys are involved.

---

### Solution

This PR adds a new Helm helper template `nd-common.overrideCommonLabels` to allow users to safely override base labels defined in `nd-common.labels`.

---

### Usage Example

Instead of appending labels manually, users can now do:

```yaml
labels:
  {{- include "nd-common.overrideCommonLabels" (dict
        "ctx" $
        "overrides" (dict
          "tags.datadoghq.com/service" (printf "nextdoor-%s-tw" $id)
          "tags.datadoghq.com/version" ($.Values.parameters.releaseToken | default "release")
        )
    ) | nindent 4 }}
```

This provides:

* Safe merging of base + user-defined labels
* Templating support for dynamic values
* Avoidance of YAML key duplication

Ref: https://github.com/Nextdoor/cloudeng/issues/201

### DIFF
--- a/charts/daemonset-app/Chart.yaml
+++ b/charts/daemonset-app/Chart.yaml
@@ -9,5 +9,5 @@ maintainers:
     email: matt@nextdoor.com
 dependencies:
   - name: nd-common
-    version: 0.4.2
+    version: 0.5.0
     repository: file://../nd-common
--- a/charts/daemonset-app/Chart.yaml
+++ b/charts/daemonset-app/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: daemonset-app
 description: Default DaemonSet Helm Chart
 type: application
-version: 0.17.2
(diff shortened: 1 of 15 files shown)


LaikaN57 (approved):
lgtm

PR #396, opened 2025-07-21
fix(charts/simple-app): make deployment zone transition label non-boolean

schahal:
Tweak to https://github.com/Nextdoor/k8s-charts/pull/391 + https://github.com/Nextdoor/k8s-charts/pull/395

K8s resources label values being unquoted boolean values can cause issues.


bwagner-nextdoor (approved):
Thanks for the fix! Figured out this was the issue late Friday but hadn't had time to put up a fix yet.

PR #402, opened 2025-07-23
feat(rollout-app): enabled seamless migration capabilities between simple-app and rollout-app

emoral435:
## Context
Currently, `simple-app` users do not have a seamless way to transition between a regular `Deployment` kind to the `Rollout` kind that is present within `rollout-app`. Luckily, the wonderful developers at Argo Rollouts have already kept this into mind while developing their CRD, and have a _built in measure_ that allows users who are using a regular deployment strategy / Deployment primitive (aka `simple-app` users). The [Rollout CRD accepts the `workloadRef` field](https://argo-rollouts.readthedocs.io/en/stable/migrating/#reference-deployment-from-rollout), which seems to be only used during migrations, so I wrapped it under a `migrate` map field. This makes it easy for developers to find within `rollout-app` (I expect developers to just be smart and ctrl+f the keyword "migrate" etc.), and makes sense to me.

Also, I ensured that users of `simple-app` who are using multiple availability zones are able to transition over as well, by ensuring that the deployment name gets changed when looping through availability zone names.

This is a big new feature, so I think that this warranted a minor version bump, but I am also open towards wrapping this up in `v1.5.x` 🙌


charts/rollout-app/templates/services.yaml
```diff
@@ -15,7 +15,7 @@ This is only used for multiple rollout deployments.
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ include "nd-common.fullname" . }}
+  name: {{ include "nd-common.serviceName" . }}
```
schahal, line 18: This is technically a breaking change, but we're allowing sync as rollout-app hasn't been used previously

schahal (approved):

PR #405, opened 2025-08-06
feat(rollout-app): add Istio subset-level traffic splitting support tweaks for multi-zone rollouts

schahal:
### 1. Support trafficRouting option of DestinationRule via subset identifiers

For more fine-grained/granular traffic routing, e.g., of an existing `Service`, if you have a `DestinationRule` that references subset identifiers on a VirtualService, Rollouts can manage that too.

https://istio.io/latest/docs/reference/config/networking/virtual-service/#Destination-subset

**So**: This change adds support of that ☝️ 

### 2. Multi-zone `Rollouts` create a `Rollout` per zone, which has Istio trafficRouting limitations

**_Multiple_** `Rollouts` objects cannot:

1.  point to (control) the same `stableService` or `canaryService` to update `VirtualService` weights.
2. point to (control) the same `DestinationRule` for subset-level traffic routing

**So**: This change ensures istio trafficRouting is not set on the `Rollout` object for multi-zone Rollouts. With the idea that as canaries come up they get added to the overall pod count along with stable at each step until promoted.

## Example Diff

values.yaml:

```yaml
virtualService:
  subsetRouting:
    enabled: true

    subsets:
      - name: stable
        weight: 100
      - name: canary
        weight: 0

    canarySubsetName: canary
    stableSubsetName: stable
    destinationRuleName: my-existing-dr
```

Rendered diff from original:

```diff
@@ -840,7 +840,14 @@
             host: rollout-app
             port:
               number: 80
+            subset: stable
           weight: 100
+        - destination:
+            host: rollout-app
+            port:
+              number: 80
+            subset: canary
+          weight: 0
 ---
 # Source: rollout-app/templates/tests/test-connection.yaml
 apiVersion: v1
@@ -3824,13 +3845,15 @@
       topology.kubernetes.io/zone: us-west-2c
   strategy:
     canary:
-      stableService: samwise
-      canaryService: samwise-canary

       trafficRouting:
         istio:
           virtualService:
             name: samwise-rollout
+          destinationRule:
+            name: samwise-app-samwise
+            canarySubsetName: canary
+            stableSubsetName: stable
       canaryMetadata:
         annotations:
           ad.datadoghq.com/tags: '{"rollout_canary": "true"}'
```


emoral435 (commented):
This change looks awesome! And I am pretty sure it should work! Just was wondering though, I looked around, and it does not seem as though `rollout-app` creates the destination rules needed to direct traffic. Perhaps looping through in a similar fashion as we do for the creation of each Rollout object for _DestinationRule_ objects would work, unless I am missing something here 🙂

LaikaN57 (approved):

